PR #12, opened 2015-05-31
Load a range of OSM objects, using a hyphen as a separator

windu2b:
Using a hyphen as a separator, range of objects IDs can be loaded.
It is easier to write "w1-5" than "w1, w2, w3, w4, w5", and the result
is exactly the same.


simon04:
Hi. This is only a mirror of the JOSM repository. Please submit patches via https://josm.openstreetmap.de/

PR #13, opened 2015-07-20
Master

Harvey-Theking:
WMTSLayer: save selected layer as default to preferences file


matkoniecz:
See description of this repository

> Git-like mirror of JOSM's Subversion repository http://josm.openstreetmap.de/

I think that it would be better to send patches via https://josm.openstreetmap.de/ (see also some closed PRs like https://github.com/openstreetmap/josm/pull/12#issuecomment-107238942)

PR #18, opened 2017-06-06
Preset search

michaelzangl:
... to allow reviews

PR #19, opened 2017-06-10
Search compiler

bafonins:
#18

### DIFF
--- a/src/org/openstreetmap/josm/actions/search/SearchAction.java
+++ b/src/org/openstreetmap/josm/actions/search/SearchAction.java
@@ -454,8 +454,8 @@ public class SearchAction extends JosmAction implements ParameterizedAction {
                 .addKeyword("type:node", "type:node ", tr("all nodes"))
                 .addKeyword("type:way", "type:way ", tr("all ways"))
                 .addKeyword("type:relation", "type:relation ", tr("all relations"))
-                .addKeyword("preset:water", "preset:water", tr("all object that match the water preset"))
-                .addKeyword("preset:\"fast food\"", "preset:\"fast food\"", tr("all objects that match the fast food preset"))
+                .addKeyword("preset:water", "preset:water", tr("all objects that use the water preset"))
+                .addKeyword("preset:\"fast food\"", "preset:\"fast food\"", tr("all objects that use the fast food preset"))
                 .addKeyword("closed", "closed ", tr("all closed ways"))
                 .addKeyword("untagged", "untagged ", tr("object without useful tags")),
                 GBC.eol());
--- a/src/org/openstreetmap/josm/actions/search/SearchAction.java
+++ b/src/org/openstreetmap/josm/actions/search/SearchAction.java
@@ -454,6 +454,8 @@ public class SearchAction extends JosmAction implements ParameterizedAction {
                 .addKeyword("type:node", "type:node ", tr("all nodes"))
                 .addKeyword("type:way", "type:way ", tr("all ways"))
                 .addKeyword("type:relation", "type:relation ", tr("all relations"))
+                .addKeyword("preset:water", "preset:water", tr("all object that match the water preset"))
+                .addKeyword("preset:\"fast food\"", "preset:\"fast food\"", tr("all objects that match the fast food preset"))
                 .addKeyword("closed", "closed ", tr("all closed ways"))
                 .addKeyword("untagged", "untagged ", tr("object without useful tags")),
                 GBC.eol());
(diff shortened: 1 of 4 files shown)


bafonins:
`    /**
     * Matches presets.
     */
    private static class Preset extends Match {
        private List<TaggingPreset> ps;

        Preset(String presetName) throws ParseError {

            if (presetName == null)
                throw new ParseError("Preset name cannot be null");

            Collection<TaggingPreset> ts = TaggingPresets.getTaggingPresets();
            ps = new ArrayList<>();

            for (TaggingPreset t : ts) {
                String name = t.getSimpleName();

                if (name != null && name.equalsIgnoreCase(presetName)) ps.add(t);
            }

            if (ps.isEmpty())
                throw new ParseError(tr("Unknown preset name: ") + presetName);
}`

src/org/openstreetmap/josm/actions/search/SearchAction.java
```diff
@@ -454,6 +454,8 @@ private static JPanel buildHintsSection(HistoryComboBox hcbSearchString) {
                 .addKeyword("type:node", "type:node ", tr("all nodes"))
                 .addKeyword("type:way", "type:way ", tr("all ways"))
                 .addKeyword("type:relation", "type:relation ", tr("all relations"))
+                .addKeyword("preset:water", "preset:water", tr("all object that match the water preset"))
```
michaelzangl: I think this would be better: "all objects that use the water preset"

src/org/openstreetmap/josm/actions/search/SearchCompiler.java
```diff
@@ -1554,6 +1558,44 @@ public String toString() {
...
 
+    /**
+     * Matches presets.
+     */
+    private static class Preset extends Match {
+        private List<TaggingPreset> ps;
+
+        Preset(String presetName) throws ParseError {
+
+            if (presetName == null)
```
michaelzangl: {

src/org/openstreetmap/josm/actions/search/SearchCompiler.java
```diff
@@ -1554,6 +1558,44 @@ public String toString() {
...
     }
 
+    /**
+     * Matches presets.
+     */
+    private static class Preset extends Match {
+        private List<TaggingPreset> ps;
+
+        Preset(String presetName) throws ParseError {
+
```
michaelzangl: <no empty line>

src/org/openstreetmap/josm/actions/search/SearchCompiler.java
```diff
@@ -1554,6 +1558,44 @@ public String toString() {
...
+            if (presetName == null)
+                throw new ParseError("Preset name cannot be null");
```
michaelzangl: "Preset name is required" (since you use this to validate the parser. If you use it to just check the parameter, use `Objects.requireNonNull`)

src/org/openstreetmap/josm/actions/search/SearchCompiler.java
```diff
@@ -1554,6 +1558,44 @@ public String toString() {
...
+
+            Collection<TaggingPreset> ts = TaggingPresets.getTaggingPresets();
+            ps = new ArrayList<>();
```
michaelzangl: You can use streams for this:
```java
ts.stream().filter(preset -> presetName.equalsIgnoreCase(preset.getSimpleName())).collect(Collectors.toList())
```

By using equals that way, you don't need the null check

Please use longer variable names.

src/org/openstreetmap/josm/actions/search/SearchCompiler.java
```diff
@@ -1554,6 +1558,44 @@ public String toString() {
...
+        }
+
+        /**
+         * Since presets can have common names, the primitive is considered to
+         * belong to a certain preset if it matches at least one of them.
+         */
+        @Override
+        public boolean match(OsmPrimitive osm) {
+            for (TaggingPreset p : ps) {
+                if (p.test(osm)) return true;
```
michaelzangl: {}

src/org/openstreetmap/josm/actions/search/SearchCompiler.java
```diff
@@ -1554,6 +1558,44 @@ public String toString() {
...
+            if (ps.isEmpty())
+                throw new ParseError(tr("Unknown preset name: ") + presetName);
+        }
+
+        /**
+         * Since presets can have common names, the primitive is considered to
+         * belong to a certain preset if it matches at least one of them.
+         */
+        @Override
+        public boolean match(OsmPrimitive osm) {
```
michaelzangl: (just a short note: Converting this to streams would not be good - streams don't have a good performance for this)

PR #22, opened 2018-01-07
 PluginHandler.installDownloadedPlugins: check downloaded plugin is valid *before* we delete any existing one

risicle:
Creating this PR in parallel to trac ticket https://josm.openstreetmap.de/ticket/15742, taking advantage of the lovely pre-merge CI here...

Noticed that the current behaviour of `PluginHandler.installDownloadedPlugins` could prove to be more than a little annoying - only checking for corruption of the newly downloaded plugin *after* it's already deleted an existing one. Someone getting repeatedly corrupted downloads or having limited internet connectivity would lose their probably-previously-working plugin.

Also added a couple of tests covering this, requiring some additions to the underlying test infrastructure, including `.assumeRevision(...)` to `JOSMTestRules` which should come in handy when testing any functionality around version-checking.


risicle:
Merged in https://josm.openstreetmap.de/changeset/13300/josm - closing.

PR #24, opened 2018-02-24
JMockit demo

risicle:
This is a patch series I'm still tidying up - taking advantage of the CI here to pre-check it...


src/org/openstreetmap/josm/tools/Logging.java
```diff
@@ -50,6 +50,66 @@
     private static final Logger LOGGER = Logger.getAnonymousLogger();
     private static final RememberWarningHandler WARNINGS = new RememberWarningHandler();
 
+    /**
+     * A {@link ConsoleHandler} with a couple of extra features, allowing it to be targeted at an
+     * an arbitrary {@link OutputStream} which it can be asked to reaquire the reference for on demand
+     * through {@link #reacquireOutputStream()}. It can also prevent a LogRecord's output if a
+     * specified {@code prioritizedHandler} would have outputted it.
+     */
+    public static class ReacquiringConsoleHandler extends ConsoleHandler {
```
don-vip: It does not seem restricted to logging. This class could be standalone.

src/org/openstreetmap/josm/data/Preferences.java
```diff
@@ -621,8 +621,14 @@ public boolean putSetting(final String key, Setting<?> setting) {
                 settingOld = settingsMap.get(key);
                 if (setting.equals(settingOld))
                     return false;
-                if (settingOld == null && setting.equals(defaultsMap.get(key)))
+                if (settingOld == null && setting.equals(defaultsMap.get(key))) {
+                    Logging.debug(tr(
```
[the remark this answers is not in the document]
risicle: Heh, this habit comes from my work project having a strict 120 char line length limit...

src/org/openstreetmap/josm/gui/util/GuiHelper.java
```diff
@@ -201,6 +201,10 @@ public static void runInEDT(Runnable task) {
         }
     }
 
+    private static void handleEDTException(Throwable t) {
+        Logging.error("Exception raised in EDT: {0}", t);
```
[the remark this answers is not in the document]
risicle: Ah only just realized that there's a specific `.error(Throwable)` which the existing logging call is using. Now I've learned of `logWithStackTrace` I'll use that.

src/org/openstreetmap/josm/tools/Logging.java
risicle: Hmm it inherits from `logging.ConsoleHandler` - I guess I could try and base it on something far more generic, but that thing would probably be so generic as to not be hugely useful and I think the resulting class would be significantly more complicated...

don-vip: OK we can keep it here then 👍

PR #29, opened 2018-08-03
Update defaultpresets.xml for tourism=artwork

vladad:


don-vip:
Material and start_date have been added through https://josm.openstreetmap.de/ticket/16602

Missing: artwork_type and artwork_subject

don-vip:
Fixed. Thanks!

PR #31, opened 2018-08-12
 test targets: add "-hardfail" variants which will abort with an error code on failure

risicle:
https://josm.openstreetmap.de/ticket/16601

Intended for use when specifying targets from top level rather than as a dependency. Also add a `test-unit` mode which will only run unit tests.

The main aim of this is to give the `.travis.yml` a target to run that will actually show up as "failed" when there were test failures. This should make it easier to drive towards a state of zero regular failures on travis.


risicle:
(merged)

PR #34, opened 2018-08-30
tests: add new tests exercising handling of plugins that advertise multiple versions for compatibility

risicle:
https://josm.openstreetmap.de/ticket/16700


risicle:
(merged)

PR #35, opened 2018-09-02
RemoteControlHttpsServer: end mainloop if server.isClosed() 

risicle:
https://josm.openstreetmap.de/ticket/16705


don-vip:
Modifications have been applied

PR #56, opened 2019-10-05
Add amenity default for scuba diving

hiddewie:
Ticket: https://josm.openstreetmap.de/ticket/16237

There should be no `sport=diving`, see https://wiki.openstreetmap.org/wiki/Tag:sport%3Ddiving


simon04:
This is only a mirror of the JOSM repository. I reported this patch in Trac as https://josm.openstreetmap.de/ticket/18539

PR #57, opened 2020-01-22
Fix relation ordering after split-way

jdhoek:
In some cases the split-way command failed to detect the direction of ways part of a relation by breaking too soon from the loop that inspects relation members. This commit fixes this.

Additionally, a warning is shown to the user if by splitting the way a relation may have been damaged. This warning is raised when all of these requirements are met:

* the relation concerned is a route or multipolygon,
* has more than one member,
* does not have all of its members downloaded, and
* the direction could not be determined from the available members.

## Background

This bug causes routes for public transport, hiking, and cycling to break when edits are performed on small parts of the map — e.g., when a mapper is working on details in a neighbourhood. It is common for such relations to have only a few members downloaded.

The split way command tries to determine the order of the new parts by looking at the available members, but due to a bug in the detection loop this can fail even when at least two connected members are available. This specific bug is hit when a way is split that is at one end of the downloaded members: in that case there is a 50% chance (depending on the position of the member in the relation) that the split will place the new parts in the wrong order. This commit solves this by not breaking from the detection loop too soon.

### DIFF
--- a/src/org/openstreetmap/josm/command/SplitWayCommand.java
+++ b/src/org/openstreetmap/josm/command/SplitWayCommand.java
@@ -53,6 +53,13 @@ public class SplitWayCommand extends SequenceCommand {
     private final List<? extends PrimitiveId> newSelection;
     private final Way originalWay;
     private final List<Way> newWays;
+
+    /**
+     * Display limit for the list of affected relations shown to the user in a warning when the order of the new way
+     * parts cannot be determined automatically.
+     */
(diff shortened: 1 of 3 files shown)


jdhoek:
The warning is shown when only one member of a relation is downloaded. For example:

warning

jdhoek:
Corresponding ticket opened in JOSM Trac: https://josm.openstreetmap.de/ticket/18596

jdhoek:
Patched upstream, available from JOSM version `r15943` onwards.

PR #58, opened 2020-04-29
Update OSMCha URL

willemarcel:


simon04:
Thanks for the patch. Since this is a Git mirror for or SVN repository, I applied the patch via https://josm.openstreetmap.de/ticket/19176 and https://josm.openstreetmap.de/changeset/16388/josm

PR #61, opened 2020-08-04
fix #19633 - Properly render links for split-start routes 

matthijskooijman:
This fixes https://josm.openstreetmap.de/ticket/19633, see the ticket for further discussion.


matthijskooijman:
I updated this PR with:
 - Some minor fixes to my original fix (amended into the original commit)
 - Added testcases of dual carriageway routes (one early commit has testcases that already worked as expected, a later commit adds some testcases for things fixed in this PR).
 - A fix for sorting split-end dual carriageway routes (fix for https://josm.openstreetmap.de/ticket/6166)
 - Some additional simplifications at the end (the last three commits). I tried to keep all refactoring small (i.e. in a few separated commits for easier review), but if you want to save some time on review, the last three commits can be left out for now too.

Should be rather complete now, there's nothing left that I want to fix before merging (that I'm aware of, anyway).

simon04:
I've applied your patch as https://josm.openstreetmap.de/changeset/16886/josm

Thank you for your work on JOSM! :-)

PR #62, opened 2020-08-05
ant: Allow running a subset of tests

matthijskooijman:
This exposes a `test-subset` target that can be used to run a subset of
tests. For example, to run a single test, specify the family and the
path to the class file (relative to the family directory):

    ant test-subset -Dtestfamily=unit -Dtestclasses="org/openstreetmap/josm/gui/dialogs/relation/sort/WayConnectionTypeCalculatorTest.class"

The `testclasses` parameter is interpreted as a pattern, so it is
usually easier to just specify the classname rather than the full path:

    ant test-subset -Dtestfamily=unit -Dtestclasses="**/WayConnectionTypeCalculatorTest.class"

Alternatively, you can run all tests from a particular directory:

    ant test-subset -Dtestfamily=unit -Dtestclasses="org/openstreetmap/josm/gui/dialogs/relation/sort/*Test.class"

Or recursively:

    ant test-subset -Dtestfamily=unit -Dtestclasses="org/openstreetmap/josm/gui/dialogs/**/*Test.class"

The console output generated by this is not ideal (A separate summary
for each test class, without including the testclass name, no overall
summary), more useful output per test can be found in the `test/report`
directory afterwards.

Ideally, more detailed info about (failing) tests would be printed to
the console directory, but it seems that ant-junitlauncher has no
builtin way to log test output to the console (junit has a
LoggerListener, but it is not directly instantiatable), so this probably
requires a custom `TestExecutionListener` class (see
https://www.selikoff.net/2018/07/28/ant-and-junit-5-outputting-test-duration-and-failure-to-the-log/
for a starting point).

### DIFF
--- a/build.xml
+++ b/build.xml
@@ -486,6 +486,12 @@ Build-Date: ${build.tstamp}
             </junitlauncher>
         </sequential>
     </macrodef>
+    <target name="test-subset" depends="test-compile" unless="test.notRequired"
+        description="Run single test class, as specified on commandline">
+        <fail message="No test family selected. Run with e.g. -Dtestfamily=unit" unless="testfamily"/>
+        <fail message="No test classes selected. Run with e.g. -Dtestclasses='**/MyTest.class'" unless="testclasses"/>
+        <call-junit testfamily="${testfamily}" includes="${testclasses}"/>
+    </target>
     <target name="test" depends="test-compile" unless="test.notRequired"
         description="Run unit and functional tests. OSM API (TEST) account shall be set with -Dosm.username and -Dosm.password">
         <call-junit testfamily="unit"/>


simon04:
Running a subset of unit tests is already possible, see https://github.com/openstreetmap/josm/commit/be40f87be2583e2a4ceeb90388a2976918905cc4:

```
$ ant test '-Ddefault-junit-includes=**/WayConnectionTypeCalculator.class'
…
test:
     [echo] Running unit tests with JUnit
[junitlauncher] 
[junitlauncher] Test run finished after 1867 ms
[junitlauncher] [         3 containers found      ]
[junitlauncher] [         0 containers skipped    ]
[junitlauncher] [         3 containers started    ]
[junitlauncher] [         0 containers aborted    ]
[junitlauncher] [         3 containers successful ]
[junitlauncher] [         0 containers failed     ]
[junitlauncher] [         6 tests found           ]
[junitlauncher] [         0 tests skipped         ]
[junitlauncher] [         6 tests started         ]
[junitlauncher] [         0 tests aborted         ]
[junitlauncher] [         6 tests successful      ]
[junitlauncher] [         0 tests failed          ]
[junitlauncher] 
     [echo] Running functional tests with JUnit

BUILD SUCCESSFUL
Total time: 6 seconds

```

matthijskooijman:
Ah, excellent point. This does not specify the test family, but just runs all of them and only one will (typically) match the include filter (or you can do `ant test-unit` to only run unit tests, no such thing for `functional`, though).

Also, this would be good to document at https://josm.openstreetmap.de/wiki/DevelopersGuide/Compiling, since that documents a somewhat involved method for running single tests that has a lot of hardcoding of .jar files and duplication in the test filename and qualified class name.

simon04:
> Also, this would be good to document at https://josm.openstreetmap.de/wiki/DevelopersGuide/Compiling

Done: https://josm.openstreetmap.de/wiki/DevelopersGuide/Compiling?action=diff&version=34

PR #63, opened 2020-09-11
#19782 changed UIAA to uiaa

mfbehrens99:
Everything was explained in the ticket


tsmock:
Ticket referenced in OP is https://josm.openstreetmap.de/ticket/19782

PR #64, opened 2020-10-05
GitHub Actions: use Ant from actions/setup-java

simon04:
Reference: https://docs.github.com/en/free-pro-team@latest/actions/guides/building-and-testing-java-with-ant


simon04:
Fixed via 99de4b020c67cbdc41589e9305ffa65bb70d9018.

PR #67, opened 2020-11-26
Improvements to macOS build and github actions

grischard:
This does a few things:

- Separate the create-release github action to pass the release url to all subsequent jobs, to be able to...
- Upload .jar artefacts from all runs, and a separate .app for all java versions for macOS
- Better caching for ant/junit
- Support for the release workflow from https://josm.openstreetmap.de/ticket/20146
- A few cosmetic fixes, more readable syntax, etc.


don-vip:
Applied in https://josm.openstreetmap.de/changeset/17373/josm, thanks a lot!

PR #69, opened 2021-02-17
Generate stable builds with ant

grischard:
Hopefully fixes https://josm.openstreetmap.de/ticket/20202

Also some drive-by comment cruft cleanup.


stoecker:
Devel is done via JOSM bug tracker, see ticket 20202

grischard:
Oh yes, those pull requests are just to conveniently create a nice linkable diff, and they can be ignored otherwise.

PR #71, opened 2021-03-18
see #20613 - Implement IQuadBucketType.intersects

simon04:
26.88% of allocations in MapCSSTagCheckerPerformanceTest.testCity amount to Way.getBBox


tsmock:
https://josm.openstreetmap.de/ticket/20613

(Also, this is open -- did you intend to reopen this?)

Anyway, it seems that `bounds(ILatLon)` is the same as `contains(ILatLon)` (in IBounds).

PR #75, opened 2021-08-25
Improve display names of route relations

matthijskooijman:
This makes two changes:
 1. Include the type of a route in its display name, to easily distinguish e.g. hiking routes from cycling routes in the UI.
 2. For objects (not necessarily limited to routes) without a name or ref, but with a from and/or to attribute, use "from-to" in the display name.
 
The initial one seems like an clear improvement. The second one might need a bit more consideration (i.e. it adds a bit of a special case).

These changes were made when working with hiking routes in the Swiss hiking route network, but should be generally usable as well.


matthijskooijman:
Is there any interest in merging this change? I'm in Switzerland again, and running into this issue again (routes that are nicely tagged with from and to, but then no longer easy to identify in josm). I'm resorting to adding note tags, but that's tagging for renderers, which isn't the best solution...

stoecker:
This is only a MIRROR! Report bugs/enhancements in https://josm.openstreetmap.de/!

src/org/openstreetmap/josm/data/osm/DefaultNameFormatter.java
```diff
@@ -410,7 +411,12 @@ public Comparator<IRelation<?>> getRelationComparator() {
 
     private static String getRelationTypeName(IRelation<?> relation) {
         // see https://josm.openstreetmap.de/browser/osm/applications/editors/josm/i18n/specialmessages.java
-        String name = trc("Relation type", relation.get("type"));
+        String type = relation.get("type");
+        String name = trc("Relation type", type);
+        String route = trc("Route type", relation.get("route"));
+        if (type != null && type.equals("route") && route != null) {
+            name = route + " " + name;
+        }
```
tsmock, line 419: `route` should be defined in the `if` statement and the `type` variable is unnecessary.

The `if` statement should probably be `relation.hasTag("type", "route") && relation.hasKey("route")`.

src/org/openstreetmap/josm/data/osm/DefaultNameFormatter.java
```diff
@@ -446,6 +452,12 @@ private static String getRelationTypeName(IRelation<?> relation) {
     private static String getNameTagValue(IRelation<?> relation, String nameTag) {
         if ("name".equals(nameTag)) {
             return formatLocalName(relation);
+        } else if ("from-to".equals(nameTag)) {
+            String from = trcLazy("from", I18n.escape(relation.get("from")));
+            String to = trcLazy("to", I18n.escape(relation.get("to")));
+            if (from != null || to != null)
+                return (from != null ? from : "?") + "-" + (to != null ? to : "?");
```
tsmock, line 459: Did you intend it to return `-<TO>` when `from` is `null` and `<FROM>-` when `to` is `null`?

tsmock (changes requested):

src/org/openstreetmap/josm/data/osm/DefaultNameFormatter.java
matthijskooijman: I usually prefer using variables like this for efficiency reasons - if you use `hasKey` and `hasTag` like your propose, you will end up doing multiple map lookups for the same key, while my code needs only one for each key. However, I can see that your proposal will result in more readable code, which is probably more important, so I'll update the code.

matthijskooijman:
First off, thanks for taking your time to respond!

> To add on to what stoecker said, once you open a ticket on https://josm.openstreetmap.de, add a link from that ticket to this PR and add `[PATCH]` to the ticket title on the JOSM website. You should also either modify the title of this PR to have the ticket number in it _or_ link back to the ticket from first post. This makes it easier to check and see if this PR has been merged, and will let someone close it if it got forgotten.

Seem I did that for an earlier PR, not sure why I did not do this here. Will do that now.

> I also haven't checked your patch -- have you run `ant pmd checkstyle` and fixed the warnings?

I'm not sure, will do so now.

> Something that I _will_ ask if you do not register on the ticket system

I registered an account, probably easier than filling in my details every time :-)

> One of the reasons why we like to have bugs/feature requests filed on our system is _just in case_ an external system decides to make life difficult for us; by having the tickets and (hopefully) most of the conversation on our own systems, we can refer back to that _even if_ the external system is closed down.

Makes sense, thanks for explaining.

> I also tend to be a bit more verbose in the commit messages/code comments since those are much less likely to be accidentally destroyed, and they tend to go with all the mirrors of the code, but that won't cover all the conversation/why something was decided. Hopefully no one is cursing my name in 10 years for poorly documenting why I did something. :)

Not sure if you looked at my commit messages yet, but I have the same tendency (this is also why I prefer a PR to patch files - easier to split changes into different commits and specify commit messages).

src/org/openstreetmap/josm/data/osm/DefaultNameFormatter.java
tsmock: That is true, but I'd rather see that from a profiling run. If there is a use case where it does have significant performance implications, avoiding the additional lookups would be fine.

But keep in mind that the JVM will do _a lot of_ optimizations, not all of which are obvious.
For example, it may generate compiled code that is _very_ similar to what you wrote.

Generally speaking, readability >>> speed. If there is a hot loop that _you have profiled_, feel free to go for speed. I've done this a few times, and I always explain why.

As an example:
```
When profiling <foo> using all the data from <area>, this was xx% of the CPU cycles and yy% of the memory allocations.
```
Examples of `<foo>`
* Validation
* Layer painting

I usually use "Mesa County, CO" as the `<area>`, since that is where I live.
One thing I will note is that memory allocations are a double whammy. They can cause the gc to perform frequent "stop the world" pauses, which stops _everything_. As an example for where reducing the memory allocations _significantly_ improves performance (as the user sees it), see https://github.com/JOSM/josm/commit/0cd6f20aa15d1d78cf436d8c7c7fe5169e62a243. In that commit, I reduced the total time spent validating a specified area by ~75% and a good chunk of that was from reducing the amount of time spent in GC by effectively removing a bunch of unnecessary memory allocations.

matthijskooijman: You already had me convinced, but thanks for following up :-)

matthijskooijman:
> I also haven't checked your patch -- have you run `ant pmd checkstyle` and fixed the warnings?

I went ahead and documented this at https://josm.openstreetmap.de/wiki/DevelopersGuide/PatchGuide?action=diff&version=12 (since I think pmd was not mentioned at all, and checkstyle only elsewhere).

matthijskooijman:
@tsmock I've made the changes you suggested, and also fixed a small formatting bug (missing space) in a new commit. The code now passes `ant checkstyle pmd test`.

When making a ticket on trac about this, I noticed an existing related ticket: https://josm.openstreetmap.de/ticket/4596#comment:9 Discussion there (from a few years ago) suggested the use of presets with name templates instead of modifying the code, which I investigated and seems a reasonable alternative to the changes in this PR, but with some significant limitations as well. See the ticket for further details, probably good to continue discussion there before further considering the code in this PR.

PR #77, opened 2021-10-27
JOSM #21480 - Add missing hints to search dialog

Woazboat:
Hints for `key?`, explicit `AND`, `XOR`


tsmock:
In [r18408](https://josm.openstreetmap.de/changeset/18408/josm)
```
fix #21480: Add missing syntax hints to search dialog (patch by Woazboat)

This adds documentation for the following syntax:
* key? (truthy values)
* &, AND
* OR
* ^, XOR (only one of)
```

PR #79, opened 2021-11-08
Fixed GPS traces no longer being browsable.

ygra:
Fixes #21538


don-vip:
Fixed in https://josm.openstreetmap.de/changeset/18320/josm thank you!

PR #80, opened 2021-11-17
Add "was" search keyword for searching in object history

Woazboat:
Added the `was` keyword to the search function to allow matching objects based on their previous states. I.e. `was <expr>` matches an object if `<expr>` matches any previous version of the object. 

Example: 
- `was building=*` to search for all objects that either currently have a `building=*` tag or have had this tag at any point in their history.
- `was building=* -building=*` to search for all objects that previously had a `building=*` tag which was removed at some point.


Woazboat:
https://josm.openstreetmap.de/ticket/21572

PR #82, opened 2021-11-28
Add style for area:highway

Woazboat:


Woazboat:
https://josm.openstreetmap.de/ticket/20102

PR #83, opened 2021-12-02
JOSM #21625 - Shrink jar/zip by further 3-5%

grischard:
5% on jar, 3% on final zip


grischard:
Trac ticket https://josm.openstreetmap.de/ticket/21625

PR #84, opened 2021-12-11
fix #21634: Download Along: Allow user to configure if there's a confirmation dialog for small number of download requests.

amandasaurus:
DownloadAlong will ask the user to confirm if it wants to make more than 1 download request. With this patch, the user can set a preference (downloadAlongWay.numDownloadRequestConfirmation or downloadAlongTrack.numDownloadRequestConfirmation), and JOSM will not ask for confirmation if it would make less than or equal to this many requests. If this preference isn't set, it's assumed to be 1. i.e. JOSM will require confirmation when it would make more than 1 request, which is the current behaviour.

I made this patch because I don't want the extra step for making (e.g.) 5 requests.

Please feel free to review it and give feedback. I'm not very experienced with java.

see also https://josm.openstreetmap.de/ticket/21634

### DIFF
--- a/src/org/openstreetmap/josm/actions/DownloadAlongAction.java
+++ b/src/org/openstreetmap/josm/actions/DownloadAlongAction.java
@@ -125,15 +125,16 @@ public abstract class DownloadAlongAction extends JosmAction {
      * @param gpxDownload Set to true if GPX data should be downloaded
      * @param title the title string for the confirmation dialog
      * @param newLayer Set to true if all areas should be put into a single new layer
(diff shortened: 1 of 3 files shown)


tsmock:
See https://josm.openstreetmap.de/ticket/21634#comment:3 for the reasoning behind closing this PR (sorry @amandasaurus ).

Again, if one of the server admins for the OSM API servers (like @firefishy) comes along and tells us that (a) the servers can handle the additional load and (b) what the maximum number of requests we can allow without verifying ought to be, I don't mind re-evaluating this.

PR #90, opened 2022-03-17
Add 'replace selected' action to relation editor

Woazboat:
Add a new action + button to the relation editor to replace chosen member primitives with objects from selection (while keeping the roles).


grischard:
Don't forget to open a ticket on the JOSM side and to link the ticket here and this PR there.

Woazboat:
Ah sorry, forgot to add the reverse link here
https://josm.openstreetmap.de/ticket/21944

src/org/openstreetmap/josm/gui/dialogs/relation/MemberTableModel.java
```diff
@@ -505,6 +505,40 @@ public void updateRole(int[] idx, String role) {
         addToSelectedMembers(selected);
     }
 
+    /**
+     * updates the referenced primitive of the members given by the index in <code>index</code>
+     *
+     * @param index the index to update
+     * @param newPrimitive the new primitive
```
tsmock, line 512:
```suggestion
     * @param newPrimitive the new primitive
     * @since xxx
```

src/org/openstreetmap/josm/gui/dialogs/relation/actions/AddFromSelectionAction.java
```diff
@@ -30,27 +30,40 @@ protected boolean isPotentialDuplicate(OsmPrimitive primitive) {
     }
 
     protected List<OsmPrimitive> filterConfirmedPrimitives(List<OsmPrimitive> primitives) throws AddAbortException {
+        return filterConfirmedPrimitives(primitives, false);
+    }
+
+    protected List<OsmPrimitive> filterConfirmedPrimitives(List<OsmPrimitive> primitives, boolean abortOnSkip) throws AddAbortException {
```
tsmock, line 36: Can you please add javadocs here?
```suggestion
    /**
     * Filter confirmed primitives
     * @param primitives The primitives to filter on
     * @param abortOnSkip If the user decides to not add a primitive, abort (throw {@code AddAbortException})
     * @return The primitives to add to the relation. Never {@code null}, but may be an empty list.
     * @throws AddAbortException when {@code abortOnSkip} is {@code true} <i>and</i> the user decides to not add a primitive.
     * @since xxx
     */
    protected List<OsmPrimitive> filterConfirmedPrimitives(List<OsmPrimitive> primitives, boolean abortOnSkip) throws AddAbortException {
```

You'll want to check this with `ant pmd checkstyle`, but it should be right.

As a side note, I'd prefer to make this method more generic (e.g. `protected <O extends IPrimitive> List<O> filterConfirmedPrimitives(List<O> primitives, boolean abortOnSkip) throws AddAbortException`), but I don't know how hard it would be to do (I suspect there would be a cascading chain of methods to change).

tsmock (changes requested):
One comment with the `replaceselectedright.svg`: As we are replacing something, should some portion of the blue bar be red (maybe two pixels or something?).

Woazboat:
I changed the icon to be partially red and fixed some checkstyle issues

PR #91, opened 2022-03-20
Fix StartupWMClass in desktop file

Cj-Malone:
Correct StartupWMClass that was causing multiple icons in the GNOME tray


tsmock:
In [r18402](https://josm.openstreetmap.de/changeset/18402/josm/).

Cj-Malone:
Thank you.

PR #92, opened 2022-03-24
Allow cardinal directions as roles in road routes.

ZLima12:
The wiki page for route relations states:
> In Canada, New Zealand, and the United States, the a role set to a cardinal direction indicates which of the route's official route directions the way carries when the route is mapped as a single bidirectional relation.

This fixes the erroneous warnings that are issued when modifying road relations mapped in this style.

### DIFF
--- a/resources/data/defaultpresets.xml
+++ b/resources/data/defaultpresets.xml
@@ -9455,6 +9455,10 @@
                     <role key="" text="route segment" requisite="optional" type="way,closedway" member_expression="highway|route=ferry"/>
                     <role key="forward" text="forward segment" requisite="optional" type="way,closedway" member_expression="highway|route=ferry"/>
                     <role key="backward" text="backward segment" requisite="optional" type="way,closedway" member_expression="highway|route=ferry"/>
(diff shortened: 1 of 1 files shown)


1ec5:
[Here’s more detail](https://wiki.openstreetmap.org/wiki/Route_directions#Cardinal_directions_as_roles) about the mapping style that this PR would accommodate. It’s one of two styles in wide use, though the PTv2-style directional subrelations are more rigorous and already supported by JOSM. A similar change was previously discussed [on the josm-dev mailing list](https://lists.openstreetmap.org/pipermail/josm-dev/2013-November/006781.html) back in 2013.

ZLima12:
I do recognize the superiorities of using one relation for each direction, but with the prevalence of this mapping style, I feel that JOSM should at least recognize it as valid. Having warnings that are "safe to dismiss" like this desensitize mappers to legitimate warnings.

ZLima12:
If we want to formally deprecate this style, then I think JOSM should give warnings about it. But until then I think it should allow it.

1ec5:
No argument there, just trying to give context about why perhaps there hadn’t been motion on this change since it was first discussed back in 2013.

ZLima12:
https://josm.openstreetmap.de/ticket/21966

PR #93, opened 2022-05-13
Drop use of abc from tile.openstreetmap.org urls

Firefishy:
Use tile.openstreetmap.org url instead of the deprecated a/b/c aliases.

Signed-off-by: Grant Slater <git@firefishy.com>

### DIFF
--- a/src/org/openstreetmap/josm/gui/bbox/JosmMapViewer.java
+++ b/src/org/openstreetmap/josm/gui/bbox/JosmMapViewer.java
@@ -91,7 +91,7 @@ public class JosmMapViewer extends JMapViewer {
     public static class DefaultOsmTileSourceProvider implements TileSourceProvider {
 
         protected static final StringProperty DEFAULT_OSM_TILE_URL = new StringProperty(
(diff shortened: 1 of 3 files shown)


tsmock:
A few comments:
* https://josm.openstreetmap.de/wiki/Maps/Worldwide#OpenStreetMapCartoStandard was updated in https://josm.openstreetmap.de/wiki/Maps/Worldwide?version=119, so we're probably OK if the a/b/c aliases stop working
* You probably didn't need to modify `src/org/openstreetmap/josm/io/remotecontrol/handler/ImageryHandler.java` (it was a usage example, so still a good idea to fix it)
* You are the person most likely to be in a position to know about the OSM tile urls, and what might magically stop working... I'm thinking the aliases are going to die soon.
* Don't count on a JOSM maintainer looking at GitHub Pull Requests. I _occasionally_ look at them, but I depend upon an RSS feed from JOSM Trac. In the future, please open a JOSM ticket and link to the GitHub PR so they are less likely to get lost.

tsmock:
In [r18452](https://josm.openstreetmap.de/changeset/18452/josm/)
```
Drop use of abc from tile.openstreetmap.org urls (patch by Firefishy)

See GitHub [#93](https://josm.openstreetmap.de/ticket/93).
```

PR #94, opened 2022-05-17
Make OH parsing a bit more lenient and update dependency

simonpoole:
This changes the behaviour of the parser to be a bit more lenient in strict mode and updates the dependency.


tsmock:
I'd close this, but I do not have `write access` to the GitHub repository.

----

In [18453/josm](https://josm.openstreetmap.de/changeset/18453/josm):

Fix [#22074](https://josm.openstreetmap.de/ticket/22074), [#22075](https://josm.openstreetmap.de/ticket/22075): OpeningHoursParser updates (patch by SimonPoole)

Disable time strict mode

    This change makes parsing in strict mode slightly more lenient with
    respect to time ranges that extend over midnight, with other words this
    will no longer complain if non-conflicting long time ranges are used
    instead of using the extended time syntax.


    Parsing in non-strict mode remains unchanged.

Update OpeningHoursParser to 0.27.0

    The update fixes an issue with parsing open ended date ranges followed
    by a weekday.

PR #95, opened 2022-05-31
JOSM #20716 - Search for missing power line support features (follow-up patch)

gabortim:
The previous way of patch sending wasn't ideal in the sense that we couldn't keep track of changes. So this way, hopefully we can save time by seeing small commits instead of the ever-growing .patch file.

### DIFF
--- a/src/org/openstreetmap/josm/data/osm/Way.java
+++ b/src/org/openstreetmap/josm/data/osm/Way.java
@@ -651,6 +651,29 @@ public final class Way extends OsmPrimitive implements IWay<Node> {
     }
 
     /**
+     * Replies the segment lengths of the way as computed by {@link LatLon#greatCircleDistance}.
+     *
+     * @return The segment lengths of a way in metres, following way direction
+     * @since xxx
+     */
+    public double[] getSegmentLengths() {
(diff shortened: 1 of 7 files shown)


tsmock:
In [18553/josm](https://josm.openstreetmap.de/changeset/18553/josm):

Fix [#20716](https://josm.openstreetmap.de/ticket/20716): Search for missing power line support features (patch by gaben, modified)

Way:
* Avoid Arrays.stream in hasIncompleteNodes. This decreases CPU cost by 90%+ and makes it free from a memory allocation standpoint.
* Add method to get segment lengths (meters) 

CrossingWays:
* Avoid Node#getEastNorth calls
* Add a getSegments call that takes ILatLon 

PowerLines:
* Check for inconsistent support node reference numbering
* Check for ways with unusually long segments without node supports
* Check for ways where line types might be misused 

ValUtil:
* Avoid unnecessary calls to Node#getEastNorth
* Add getSegmentCells for ILatLon 

InspectPrimitiveDataText:
* Add average segment length statistic 

Geometry:
* Add getSegmentSegmentIntersection for ILatLon to avoid new EastNorth objects 

Utils:
* Add getStandardDeviation methods

PR #100, opened 2022-08-23
Add option to merge by default when dragging nodes

Woazboat:
i.e. merge dragged nodes when they are moved over other nodes without having to hold ctrl

Holding ctrl while moving a node around toggles merge mode like before
(i.e. enables when off by default, disables when on by default)

### DIFF
--- a/src/org/openstreetmap/josm/actions/mapmode/SelectAction.java
+++ b/src/org/openstreetmap/josm/actions/mapmode/SelectAction.java
@@ -42,6 +42,8 @@ import org.openstreetmap.josm.data.osm.Way;
 import org.openstreetmap.josm.data.osm.WaySegment;
 import org.openstreetmap.josm.data.osm.visitor.AllNodesVisitor;
 import org.openstreetmap.josm.data.osm.visitor.paint.WireframeMapRenderer;
+import org.openstreetmap.josm.data.preferences.BooleanProperty;
+import org.openstreetmap.josm.data.preferences.CachingProperty;
 import org.openstreetmap.josm.gui.ExtendedDialog;
 import org.openstreetmap.josm.gui.MainApplication;
 import org.openstreetmap.josm.gui.MapFrame;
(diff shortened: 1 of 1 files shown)


src/org/openstreetmap/josm/actions/mapmode/SelectAction.java
```diff
@@ -131,6 +133,10 @@ public Cursor cursor() {
         }
     }
 
+    /** Whether nodes should be merged with other primitives by default when they are being dragged */
+    static final CachingProperty<Boolean> MERGE_BY_DEFAULT
+            = new BooleanProperty("edit.move.merge-by-default", false).cached();
```
tsmock, line 138: Thank you for using a cached property. I presume you did some profiling and noticed that not caching it caused lag. :)

Also, it makes sense for it to be `false` by default to conform with current behavior.

Stupid question:
Have you considered adding it as a configuration option in JOSM Preferences -> `Display` (just above `Reverse zoom with mouse wheel` perhaps?)

Woazboat: I didn't actually profile it (or notice lag) because I never tried it non-cached, but I don't think there's a reason not to (?) and it definitely makes sense here.

> Also, it makes sense for it to be false by default to conform with current behavior.

Yes, I definitely wanted to keep the current behaviour as default. I think merging objects should be deliberate and having it on by default is a little bit more likely to cause unintended and potentially unnoticed accidents.

> Have you considered adding it as a configuration option in JOSM Preferences -> Display (just above Reverse zoom with mouse wheel perhaps?)

I wanted to add it to the preferences dialog but then I noticed there's actually no section for options that influence editing behavior. The sections that currently exist are only about visual appearance and I don't think it should be shoehorned in there. This is something that should be done in a future patch.

tsmock: One of the reasons to avoid caching properties is to avoid having a bunch of references lying around in memory. Not a major issue, unless you are caching/creating properties in a while loop.

As far as preferences go, it is probably fine for now, but I worry about forgetting stuff.

tsmock:
In [18554/josm](https://josm.openstreetmap.de/changeset/18554/josm):

Fix [#22309](https://josm.openstreetmap.de/ticket/22309): Add option to merge by default when dragging nodes (patch by Woazboat)

This is currently an advanced preference only. It uses the ctrl key to manage
merging.

PR #103, opened 2022-08-31
JOSM #22333 - Allow delete + move up/down actions in filter dialog to act on all selected filters

Woazboat:
The `delete` as well as the `move up`/`move down` actions in the filter dialog previously only acted on the first filter in the selection, even if multiple filters were selected simultaneously. With this patch these actions act on all selected filters simultaneously.

https://josm.openstreetmap.de/ticket/22333

### DIFF
--- a/src/org/openstreetmap/josm/gui/dialogs/FilterDialog.java
+++ b/src/org/openstreetmap/josm/gui/dialogs/FilterDialog.java
@@ -9,6 +9,7 @@ import java.awt.event.ActionEvent;
 import java.awt.event.KeyEvent;
 import java.awt.event.MouseEvent;
 import java.util.Arrays;
+import java.util.Collections;
 import java.util.List;
 import java.util.stream.Collectors;
 import java.util.stream.IntStream;
@@ -175,10 +176,11 @@ public class FilterDialog extends ToggleDialog implements DataSetListener, MapMo
 
         @Override
         public void actionPerformed(ActionEvent e) {
-            int index = filterModel.getSelectionModel().getMinSelectionIndex();
-            if (index >= 0) {
-                filterModel.removeFilter(index);
-            }
+            // Remove in reverse order to preserve index validity
+            int[] selected = filterModel.getSelectedIndices();
+            Arrays.stream(selected).boxed()
+                .sorted(Collections.reverseOrder())
+                .forEachOrdered(filterModel::removeFilter);
         }
     }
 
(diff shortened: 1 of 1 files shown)


src/org/openstreetmap/josm/gui/dialogs/FilterDialog.java
```diff
@@ -175,10 +176,11 @@ private class DeleteAction extends FilterAction {
...
         public void actionPerformed(ActionEvent e) {
-            int index = filterModel.getSelectionModel().getMinSelectionIndex();
-            if (index >= 0) {
-                filterModel.removeFilter(index);
-            }
+            // Remove in reverse order to preserve index validity
+            int[] selected = filterModel.getSelectedIndices();
+            Arrays.stream(selected).boxed()
+                .sorted(Collections.reverseOrder())
+                .forEachOrdered(filterModel::removeFilter);
```
tsmock, line 183: I really don't like this code.

FilterTableModel#fireTableRowsDeleted calls `savePrefs()`, `updateFilters()`, and `fireTableRowsDeleted` _every time_. This can get expensive quickly.

I'd much prefer to have a `removeFilters(int... rowIndexes)` method in FilterTableModel, which would look something like this:
```java
    /**
     * Removes the filters that is displayed in the given rows
     * @param rowIndex The index of the filters to remove
     * @since xxx
     */
    public void removeFilters(int... rowIndex) {
        boolean modified = false;
        Arrays.sort(rowIndex);
        for (int i = rowIndex.length - 1; i >= 0; i--) {
            modified |= model.removeFilter(i) != null;
        }
        if (modified) {
            savePrefs();
            updateFilters();
            List<int[]> groupedRows = groupRows(rowIndex);
            // Reverse to avoid having to deal with offsets
            Collections.reverse(groupedRows);
            for (int[] rows : groupedRows) {
                fireTableRowsDeleted(rows[0], rows[1]);
            }
        }
    }
```

Note: I added `groupRows` in https://josm.openstreetmap.de/changeset/18439/josm/, but it isn't available. It should probably be moved to Utils and renamed.

tsmock (changes requested):
I'm going to need to look into why we aren't using filterModel.moveUp/moveDown

tsmock:
It looks like the code hasn't been substantially changed except as part of larger refactorings since it was first introduced ( https://josm.openstreetmap.de/browser/josm/trunk/src/org/openstreetmap/josm/gui/dialogs/FilterDialog.java?annotate=blame&rev=2125#L114 ). We should absolutely use the (much) newer filterModel.moveUp/moveDown methods.

tsmock:
In [18556/josm](https://josm.openstreetmap.de/changeset/18556/josm):

Fix [#22333](https://josm.openstreetmap.de/ticket/22333): Allow delete + move up/down actions in filter dialog to act on all selected filters (patch by Woazboat, modified)

This also adds build add/remove methods
in FilterModel and FilterTableModel.

PR #104, opened 2022-09-16
JOSM #22378 - Preview object properties on mouse hover

Woazboat:
https://josm.openstreetmap.de/ticket/22378

Show tags and relation memberships of an object in the properties dialog when moving the mouser pointer over it (similar to iD).

Can be enabled/disabled via settings (enabled by default)

An existing selection overrides the hover preview by default, but this behavior can be changed via a setting to always show information about hovered objects.

josm_mouse_hover_object_preview

### DIFF
--- a/src/org/openstreetmap/josm/gui/NavigatableComponent.java
+++ b/src/org/openstreetmap/josm/gui/NavigatableComponent.java
@@ -8,6 +8,8 @@ import java.awt.event.ComponentAdapter;
 import java.awt.event.ComponentEvent;
 import java.awt.event.HierarchyEvent;
 import java.awt.event.HierarchyListener;
+import java.awt.event.MouseAdapter;
+import java.awt.event.MouseEvent;
 import java.awt.geom.AffineTransform;
 import java.awt.geom.Point2D;
 import java.nio.charset.StandardCharsets;
@@ -20,6 +22,7 @@ import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
+import java.util.Objects;
 import java.util.Set;
 import java.util.Stack;
 import java.util.TreeMap;
@@ -52,6 +55,7 @@ import org.openstreetmap.josm.data.preferences.IntegerProperty;
 import org.openstreetmap.josm.data.projection.Projection;
 import org.openstreetmap.josm.data.projection.ProjectionChangeListener;
 import org.openstreetmap.josm.data.projection.ProjectionRegistry;
+import org.openstreetmap.josm.gui.PrimitiveHoverListener.PrimitiveHoverEvent;
 import org.openstreetmap.josm.gui.help.Helpful;
 import org.openstreetmap.josm.gui.layer.NativeScaleLayer;
 import org.openstreetmap.josm.gui.layer.NativeScaleLayer.Scale;
@@ -63,6 +67,7 @@ import org.openstreetmap.josm.gui.util.GuiHelper;
 import org.openstreetmap.josm.spi.preferences.Config;
 import org.openstreetmap.josm.tools.Logging;
 import org.openstreetmap.josm.tools.Utils;
+import org.openstreetmap.josm.tools.bugreport.BugReportExceptionHandler;
 
 /**
  * A component that can be navigated by a {@link MapMover}. Used as map view and for the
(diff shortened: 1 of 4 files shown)


Woazboat:
Object selection state could be cached so that ui updates are only triggered when there are actual changes. That's not in there at the moment but could be added if there are performance problems.

src/org/openstreetmap/josm/gui/dialogs/properties/PropertiesDialog.java
```diff
@@ -635,7 +671,37 @@ public void selectionChanged(SelectionChangeEvent event) {
...
+
+        List<OsmPrimitive> hovered = new ArrayList<>();
+        p.ifPresent(hovered::add);
+        updateUi(hovered);
+    }
+
+    @Override
+    public void mouseDragged(MouseEvent e) {
+        // Do nothing
+    }
```
tsmock, line 695: I'm really not a fan of this, but we don't have a better option in core right now.

I'd like to have a `HighlightManager` or something which passes highlight/hover information around, but we don't have that yet. :(

src/org/openstreetmap/josm/gui/dialogs/properties/PropertiesDialog.java
```diff
@@ -304,6 +316,21 @@ public PropertiesDialog() {
...
+        @Override
+        public void valueChanged(ValueChangeEvent<? extends Boolean> e) {
+            if (e.getProperty().get() && isDialogShowing() && !haveMouseMotionListener) {
+                MainApplication.getMap().mapView.addMouseMotionListener(PropertiesDialog.this);
+                haveMouseMotionListener = true;
+            } else if (!e.getProperty().get() && haveMouseMotionListener) {
+                MainApplication.getMap().mapView.removeMouseMotionListener(PropertiesDialog.this);
+                haveMouseMotionListener = false;
+            }
+        }
```
tsmock, line 333: This should probably be down near the bottom of the file, with the other private class.

src/org/openstreetmap/josm/gui/dialogs/properties/PropertiesDialog.java
```diff
@@ -679,14 +772,18 @@ public void selectionChanged(SelectionChangeEvent event) {
...
 
+        presets.updatePresets(types, tags, presetHandler);
+    }
+
+    private void updateMembershipTableData(Collection<? extends IPrimitive> primitives) {
         membershipData.setRowCount(0);
 
         Map<IRelation<?>, MemberInfo> roles = new HashMap<>();
-        for (IPrimitive primitive: newSel) {
+        for (IPrimitive primitive: primitives) {
```
tsmock, line 782:
```suggestion
        for (IPrimitive primitive : primitives) {
```
Since you modified this line. :)

tsmock (changes requested):
Overall, I think this looks good. I didn't see any significant performance degradation during my testing, but there is some optimization that can occur. I would be more inclined to fix the problem in TaggingPreset#isShowable instead of in your code though.

It might still be worth it to check and see if the selection hasn't changed, but I don't expect it to make a difference.

src/org/openstreetmap/josm/gui/dialogs/properties/PropertiesDialog.java
Woazboat: I originally put it up there to be close to where it's actually being used, but I moved it to the bottom now.

src/org/openstreetmap/josm/gui/dialogs/properties/PropertiesDialog.java
Woazboat: Yeah same, it's not the prettiest piece of code but the only real option at the moment. I might look into adding a `HoverListener` to the map view that client code can use to be notified whenever the object under the mouse pointer changes. That could be reused in all sorts of places.

Woazboat:
I just noticed a minor issue: The preview is also displayed and updated when the data layer is hidden and the hovered objects are not actually visible.

tsmock:
Take a look at Layer.VISIBLE_PROP in conjunction with implementing a PropertyChangeListener.

Example:
```java
    @Override
    public void activeOrEditLayerChanged(ActiveLayerChangeEvent e) {
        // Remove previous listener (order matters if we are somehow getting a layer change event switching from one layer to the same layer)
        if (e.getPreviousDataLayer() != null) {
            e.getPreviousDataLayer().removePropertyChangeListener(this);
        }
        // Register a listener
        if (e.getSource().getEditLayer() != null) {
            e.getSource().getEditLayer().addPropertyChangeListener(this);
        }
    }
    @Override
    public void propertyChange(PropertyChangeEvent evt) {
        if (Layer.VISIBLE_PROP.equals(evt.getPropertyName())) {
            boolean isVisible = (boolean) evt.getNewValue();
            // Update state
        }
    }
```

src/org/openstreetmap/josm/gui/NavigatableComponent.java
```diff
@@ -149,6 +154,51 @@ protected static void fireZoomChanged() {
         });
     }
 
+    private final CopyOnWriteArrayList<PrimitiveHoverListener> primitiveHoverListeners = new CopyOnWriteArrayList<>();
+    // Allow primitive to be garbage collected by using a weak reference
+    private WeakReference<OsmPrimitive> previousHoveredPrimitive = new WeakReference<>(null);
```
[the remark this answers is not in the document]
Woazboat: Wasn't entirely sure if it would be worth it to do it this way but I added it to be on the safe side. I don't know if there are some edge cases where primitives can get really large, but I guess those really are edge cases and it shouldn't stick around for long in any case. Will change it back to a normal variable.

> If code users want to keep track of the previous object, that should probably be done in their code.

We need to keep track of the previous object anyway so that we can fire the hover events only when something changed.

src/org/openstreetmap/josm/gui/NavigatableComponent.java
```diff
@@ -149,6 +154,51 @@ protected static void fireZoomChanged() {
...
+    }
+
+    protected void firePrimitiveHovered(PrimitiveHoverEvent e) {
+        GuiHelper.runInEDT(() -> {
+            for (PrimitiveHoverListener l : primitiveHoverListeners) {
+                try {
+                    l.primitiveHovered(e);
+                } catch (Exception ex) {
+                    Logging.error(ex);
+                }
```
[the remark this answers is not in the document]
Woazboat: Exceptions are caught here to avoid a situation like this https://josm.openstreetmap.de/ticket/21682
I expect that quite a few plugins might want to use this listener and an error in a plugin should not interfere with the operation of the rest of the application.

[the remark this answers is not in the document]
tsmock: Both true and false. I'd rather force plugins to fix their code than have plugins _think_ something is working when it isn't. How often do people actually read all the output from the `Logging` methods?

If it turns out to block JOSM shutdown, we can re-evaluate.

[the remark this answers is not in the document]
Woazboat: I changed it to pop the bug report window (which is what I wanted to do in the first place but somehow thought the error logging already did)

Woazboat:
I added a setting that controls whether a selection should override the hover preview (default is to always show the selection).

Woazboat:
Are there any other remarks from your side? Otherwise I think this is good to go.

tsmock:
Good to know. I'll take a look at it again today. I won't merge it until Thursday at the earliest (we did a hotfix last Friday, and I'm waiting to make certain that we don't have any other bugs that affect many users, as counted by duplicate tickets).

tsmock:
In [18574/josm](https://josm.openstreetmap.de/changeset/18574/josm):

Fix [#22378](https://josm.openstreetmap.de/ticket/22378): Preview object properties on mouse hover (patch by Woazboat, modified)

Show tags and relation memberships of an object in the properties dialog when
moving the mouser pointer over it (similar to iD).

Can be enabled/disabled via settings (enabled by default)

PR #105, opened 2022-11-05
JOSM #22487 - Fix hover preview listener being incorrectly re-added on layer change

Woazboat:
Layer (visibility) change listeners did not check if hover preview is disabled and always added the primitive hover listener even if they shouldn't, effectively re-enabling hover preview on layer (visibility) change.

[#22487](https://josm.openstreetmap.de/ticket/22487)


tsmock:
In [18586/josm](https://josm.openstreetmap.de/changeset/18586/josm):

Fix [#22487](https://josm.openstreetmap.de/ticket/22487): Object hover preview is re-enabled on layer change even when disabled in properties (patch by Woazboat, modified)

PR #107, opened 2023-01-11
Fix hover preview setting changes not being applied immediately

Woazboat:
Hover listener was not re-added correctly when prioritize selection was turned off. (Only caused issues if the hover listener was currently disabled due to a non-empty selection)


tsmock:
In [18631/josm](https://josm.openstreetmap.de/changeset/18631/josm):

See [#22378](https://josm.openstreetmap.de/ticket/22378): Fix hover preview setting changes not being applied immediately

PR #109, opened 2023-02-03
JOSM #22703 - Fix mapcss rotation offset for *-rotation: way

Woazboat:
https://josm.openstreetmap.de/ticket/22703

The (undocumented)  `*-rotation: way;`in mapcss has an offset of 90° compared to `*-rotation: <fixed angle>`

The 90 degree offset occurs because https://josm.openstreetmap.de/browser/josm/trunk/src/org/openstreetmap/josm/tools/RotationAngle.java#L37 uses Geometry.getSegmentAngle, which calculates the angle using Math.atan2 and therefore the rotation starting from the x-axis (east) in counter clockwise direction. The icon rotation is relative to the icon and therefore 'starts' at the y-axis (north) going in the clockwise direction.
The returned angle is already correctly inverted to fix the rotation angle, but the 90° offset is not corrected. This can be fixed by adding an offset of - Math.PI/2 to the angle (+Math.PI/2 if done after inversion).

### DIFF
--- a/src/org/openstreetmap/josm/tools/RotationAngle.java
+++ b/src/org/openstreetmap/josm/tools/RotationAngle.java
@@ -32,11 +32,27 @@ public interface RotationAngle {
                 return 0;
             }
             final Way w = ways.iterator().next();
+            Double angle = getRotationAngleForNodeOnWay(n, w);
+            return angle != null ? angle : 0;
+        }
+
+        /**
+         * Calculates the rotation angle of a node in a way based on the preceding way segment.
+         * If the node is the first node in the given way, the angle of the following way segment is used instead.
+         * @param n the node to get the rotation angle for
+         * @param w the way containing the node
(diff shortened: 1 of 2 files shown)


tsmock:
In [18661/josm](https://josm.openstreetmap.de/changeset/18661/josm):

Fix [#22703](https://josm.openstreetmap.de/ticket/22703), see [#22695](https://josm.openstreetmap.de/ticket/22695), [#22704](https://josm.openstreetmap.de/ticket/22704): 90° offset of icon-rotation: way; in mapcss (patch by Woazboat, modified)

This is due to Geometry.getSegmentAngle getting the rotation from the positive
x-axis instead of the positive y-axis, but this is easily corrected by
subtracting 90 degrees (pi/2) from the calculation.

The previous behavior differed from the static rotation angle behavior, which
uses the positive y-axis.

PR #110, opened 2023-02-03
JOSM #22704 - Add `parent_way_angle()` mapcss function

Woazboat:
https://josm.openstreetmap.de/ticket/22704

Based on https://github.com/JOSM/josm/pull/109

Add a mapcss function that returns the angle of the parent way section of a node (matched via child element selector) in radians. Similar to the existing `icon-rotation: way;`, just as a function which enables a lot more flexibility.


tsmock:
In [18661/josm](https://josm.openstreetmap.de/changeset/18661/josm):

Fix [#22703](https://josm.openstreetmap.de/ticket/22703), see [#22695](https://josm.openstreetmap.de/ticket/22695), [#22704](https://josm.openstreetmap.de/ticket/22704): 90° offset of icon-rotation: way; in mapcss (patch by Woazboat, modified)

This is due to Geometry.getSegmentAngle getting the rotation from the positive
x-axis instead of the positive y-axis, but this is easily corrected by
subtracting 90 degrees (pi/2) from the calculation.

The previous behavior differed from the static rotation angle behavior, which
uses the positive y-axis.

Woazboat:
rebased onto trunk

tsmock:
In [18664/josm](https://josm.openstreetmap.de/changeset/18664/josm):

Fix [#22704](https://josm.openstreetmap.de/ticket/22704): Add parent_way_angle mapcss function (patch by Woazboat)

PR #111, opened 2023-02-08
JOSM 22716: Remove duplicate line

Famlam:
See https://josm.openstreetmap.de/ticket/22716
Line 66 and 73 both add a warning for `natural=wood` on a node


tsmock:
In [18653/josm](https://josm.openstreetmap.de/changeset/18653/josm):

Fix [#22716](https://josm.openstreetmap.de/ticket/22716): Duplicate natural=wood line in geometry.mapcss caused duplicate warnings (patch by Famlam, modified)

This patch was modified to also sort the natural tag section of the check, so
that it is harder to miss duplicates in the future.

PR #112, opened 2023-02-13
Fix for 21778: bad warning if oneway:conditional (or oneway:*:conditional) is present

Famlam:
This fixes the mapcss part of https://josm.openstreetmap.de/ticket/21778
Since the Java part has (as the comment by Skyper implies) more issues, I left that untouched for now, even though the fix is very easy.


tsmock:
In [18674/josm](https://josm.openstreetmap.de/changeset/18674/josm):

Fix [#21778](https://josm.openstreetmap.de/ticket/21778): oneway + .*:(backward|forward):conditional should not have a warning (patch by Famlam)

PR #113, opened 2023-02-15
fixed typo

Sarabjeet108:


tsmock:
Heads up: We don't typically apply typo-only fixes, unless the typo is in the UI and could confuse users, or is in the code (for example, `varargMethod(MessageFormater.format("{0}"), "Hello world")`).

You should also know that this project is developed in `svn`, which means that whoever commits your change will have to make a commit message like `fixed typo (patch by Sarabjeet108)`. This repository is a SVN mirror.

With that said, we do welcome patches. In the future, please at least _refer_ to the PR in a ticket on [JOSM Trac](josm.openstreetmap.de).

PR #115, opened 2023-03-14
Use twelvemonkeys.imageio and add webp support

Firefishy:
WARNING: I am not a java developer... That said, this seems to work well for me.

Switching to twelvemonkeys imageio and adding their imageio-webp module was all that was required for me to enable webp tile support.

I have tested png, jpeg using both tile and WMS and they continue to work as expected (mapbox, osm carto, maxar...)

### DIFF
--- a/ivy.xml
+++ b/ivy.xml
@@ -20,6 +20,14 @@
     </configurations>
     <dependencies>
         <!-- api->default -->
+        <dependency conf="api->default" org="com.twelvemonkeys.common" name="common-lang" rev="3.9.4"/>
+        <dependency conf="api->default" org="com.twelvemonkeys.common" name="common-io" rev="3.9.4"/>
+        <dependency conf="api->default" org="com.twelvemonkeys.common" name="common-image" rev="3.9.4"/>
+        <dependency conf="api->default" org="com.twelvemonkeys.imageio" name="imageio-core" rev="3.9.4"/>
+        <dependency conf="api->default" org="com.twelvemonkeys.imageio" name="imageio-metadata" rev="3.9.4"/>
+        <dependency conf="api->default" org="com.twelvemonkeys.imageio" name="imageio-jpeg" rev="3.9.4"/>
+        <dependency conf="api->default" org="com.twelvemonkeys.imageio" name="imageio-tiff" rev="3.9.4"/>
+        <dependency conf="api->default" org="com.twelvemonkeys.imageio" name="imageio-webp" rev="3.9.4"/>
         <dependency conf="api->default" org="org.openstreetmap.jmapviewer" name="jmapviewer" rev="2.16"/>
         <dependency conf="api->default" org="javax.json" name="javax.json-api" rev="1.1.4"/>
         <dependency conf="api->default" org="org.glassfish" name="javax.json" rev="1.1.4"/>


tsmock:
For reference, this appears to increase jar size by 757523 bytes (~4% increase). I do know I did meet resistance when I wanted to add Google's protobuf library (1676712 bytes, ~10% increase), I encountered resistance and just ended up writing a JOSM internal PBF parser.

stoecker:
Sounds more like "for a plugin". E.g. opendata, dataimport or a new "imageformats" plugin.

Firefishy:
My interest in webp is because I've been investigating switching many openstreetmap.org tile services to webp format, as it is 25% smaller and would save significant bandwidth and improve performance. It is now widely supported by browsers https://caniuse.com/webp

tsmock:
In that case, it _might_ be worth it. It would be an easier sell if it wasn't bundled into JOSM (as an example, an automatically enabled plugin would probably be fine, with the caveat that it would _have to be_ controlled by the JOSM maintainers).

Using https://tile.openstreetmap.org/16/13045/25020.png for my test, optipng to optimize the png, and cwebp for conversions, I got the following results:

| | PNG | PNG (optimized) | WebP (q100) | WebP (q90) | WebP (q80) |
|-------|-------|------|---------------|-----------|------------|
| Bytes | 1918 | 1615 | 1874 | 1198 | 912 |
| Difference | 0 | -303 | -44 | -720 | -1006 |
| % Difference | 0% | -15.8% | -2.3% | -37.5% | -52.5% |

I assume you (@Firefishy) did some more in-depth tests for the 25% smaller number. Figure ~500 bytes saved per tile, this means that each JOSM user will have to download ~1500 (757523/500) tiles per month to make an overall bandwidth difference for the user. Assuming the png files are not optimized.

Still, for something like this, it would probably be better to have a plugin which manages the ImageIO plugins and adds them to the classpath/module path. That would be a much more manageable download hit for the users, with the added benefit that it can be updated out-of-band (just in case there is a significant security vulnerability). Ideally the plugin would check the maven repo for updated dependencies and automatically download them as needed.

I'll try to remember to look into making a plugin for it once I go into freeze mode for the March release (around March 22nd).

tsmock:
I'll note that running the png file through `optipng` showed a 15% decrease, which makes WebP at q100 even worse. I _did_ specify the programs and parameters used just in case there were better options.

Anyway, I just ran `cwebp 25020.png -lossless -o 25020.webp` and that was 1146 bytes. I would have assumed (wrongly) that `q100` would be equivalent to `-lossless` (which didn't show up in the initial `cwebp -h` options).

Firefishy:
PNG files created by all OpenStreetMap.org imagery services are generated on demand and cannot be optimised with `optipng` or similar.

stoecker:
> FYI: PNG files created by all OpenStreetMap.org imagery services are generated on demand and cannot be optimised with `optipng` or similar.

Are you sure? They possibly could be optimized in two places:
* in the mapnik(?) generating chain
* by a background job optimizing the cached data files.

Not as straightforward as calling optipng on a PNG file, but possible.

tsmock:
This has been added as a plugin (`ImageIO` (JOSM svn) and `webp` (GitHub, not JOSM controlled)). The `webp` plugin apparently has issues with ARM macs. I don't _think_ the `ImageIO` plugin will, since the library it uses doesn't have native code.

----

In [36094/osm](https://josm.openstreetmap.de/changeset/36094/osm):

See [#14361](https://josm.openstreetmap.de/ticket/14361) (dist): Support for image/webp

PR #117, opened 2023-03-23
Fixes JOSM#22546 - add `railway=switch` & `railway=railway_crossing`

Famlam:
- Adds `railway=switch` and `railway=railway_crossing`
- Sorts the railway values alphabetically

https://josm.openstreetmap.de/ticket/22546


tsmock:
In [18914/josm](https://josm.openstreetmap.de/changeset/18914/josm):

Fix [#22546](https://josm.openstreetmap.de/ticket/22546): Warn about floating railway switches and crossing (patch by Famlam)

railway=switch and railway=crossing should always be connected to a railway;
neither can exist without a railway line as they are physical features of the
railway.

PR #120, opened 2023-05-12
Improve description wording

simonpoole:
It would seem that this is being undersold relative to iD.


tsmock:
In [https://josm.openstreetmap.de/changeset/18760/josm/ r18760].

PR #122, opened 2023-06-22
Add support for regions attribute

Sarabjeet108:
@tsmock

### DIFF
--- a/resources/data/defaultpresets.xml
+++ b/resources/data/defaultpresets.xml
@@ -2844,7 +2844,7 @@
             <reference ref="link_contact_address_payment" />
             <preset_link preset_name="Building" />
         </item> <!-- Fuel -->
-        <item name="Charging Station" icon="presets/vehicle/charging_station.svg" regions="US,CA" type="node" preset_name_label="true">
+        <item name="Charging Station" icon="presets/vehicle/charging_station.svg" type="node" preset_name_label="true">
             <link wiki="Tag:amenity=charging_station" />
             <space />
             <key key="amenity" value="charging_station" />
--- a/resources/data/defaultpresets.xml
+++ b/resources/data/defaultpresets.xml
@@ -2844,7 +2844,7 @@
             <reference ref="link_contact_address_payment" />
             <preset_link preset_name="Building" />
         </item> <!-- Fuel -->
-        <item name="Charging Station" icon="presets/vehicle/charging_station.svg" type="node" preset_name_label="true">
+        <item name="Charging Station" icon="presets/vehicle/charging_station.svg" regions="US,CA" type="node" preset_name_label="true">
             <link wiki="Tag:amenity=charging_station" />
             <space />
             <key key="amenity" value="charging_station" />
--- a/resources/data/tagging-preset.xsd
+++ b/resources/data/tagging-preset.xsd
@@ -121,7 +121,7 @@
     <complexType name="item">
         <annotation>
             <documentation>
-                Every item is one annotation set to select from. name is required, type and preset_name_label are recommended, icon and name_template are optional attributes.
+                Every item is one annotation set to select from. name is required, type and preset_name_label are recommended and icon, name_template, regions and exclude_regions are optional attributes.
             </documentation>
         </annotation>
         <sequence>
@@ -168,6 +168,20 @@
             <annotation>
                 <documentation>
                     Whether to insert "[preset name]" label in dialog (defaults to "false" for backward-compatibility on external presets).
+                </documentation>
+            </annotation>
+        </attribute>
+        <attribute name="regions" type="string">
+            <annotation>
+                <documentation>
+                    Comma separated list of countries this preset group or item is applicable for. If not specified, the preset is applicable for all countries.
+                </documentation>
+            </annotation>
+        </attribute>
+        <attribute name="exclude_regions" type="boolean">
+            <annotation>
+                <documentation>
+                    If true, invert the meaning of regions.
                 </documentation>
             </annotation>
         </attribute>
--- a/src/org/openstreetmap/josm/data/validation/tests/TagChecker.java
+++ b/src/org/openstreetmap/josm/data/validation/tests/TagChecker.java
@@ -664,14 +664,16 @@ public class TagChecker extends TagTest implements TaggingPresetListener {
         }
 
         final Collection<TaggingPreset> matchingPresets;
-        TagMap tags = p.getKeys();
+        TagMap tags;
         if (checkPresetsTypes || checkRegions) {
+            tags = p.getKeys();
             matchingPresets = presetIndex.entrySet().stream()
                     .filter(e -> TaggingPresetItem.matches(e.getValue(), tags))
                     .map(Entry::getKey)
                     .collect(Collectors.toCollection(LinkedHashSet::new));
         } else {
             matchingPresets = null;
+            tags = null;
         }
 
         if (checkPresetsTypes) {
@@ -721,7 +723,7 @@ public class TagChecker extends TagTest implements TaggingPresetListener {
                     }
                     if (isInRegion == preset.exclude_regions()) {
                         errors.add(TestError.builder(this, Severity.WARNING, INVALID_REGION)
-                                .message(tr("Preset not valid for region"),
+                                .message(tr("Invalid region for this preset"),
                                         marktr("Preset {0} should not be applied in this region"),
                                         tr(preset.getName()))
                                 .primitives(p)
--- a/src/org/openstreetmap/josm/data/validation/tests/TagChecker.java
+++ b/src/org/openstreetmap/josm/data/validation/tests/TagChecker.java
@@ -4,6 +4,7 @@ package org.openstreetmap.josm.data.validation.tests;
 import static org.openstreetmap.josm.tools.I18n.marktr;
 import static org.openstreetmap.josm.tools.I18n.tr;
 import static org.openstreetmap.josm.tools.I18n.trn;
+import static org.openstreetmap.josm.tools.Territories.isIso3166Code;
 
 import java.awt.GridBagConstraints;
 import java.awt.event.ActionListener;
@@ -37,8 +38,10 @@ import org.openstreetmap.josm.command.ChangePropertyCommand;
 import org.openstreetmap.josm.command.ChangePropertyKeyCommand;
 import org.openstreetmap.josm.command.Command;
 import org.openstreetmap.josm.command.SequenceCommand;
+import org.openstreetmap.josm.data.coor.LatLon;
 import org.openstreetmap.josm.data.osm.AbstractPrimitive;
 import org.openstreetmap.josm.data.osm.DataSet;
+import org.openstreetmap.josm.data.osm.Node;
 import org.openstreetmap.josm.data.osm.OsmPrimitive;
 import org.openstreetmap.josm.data.osm.OsmUtils;
 import org.openstreetmap.josm.data.osm.Relation;
@@ -131,6 +134,7 @@ public class TagChecker extends TagTest implements TaggingPresetListener {
      * The preference key to check presets
      */
     public static final String PREF_CHECK_PRESETS_TYPES = PREFIX + ".checkPresetsTypes";
+    public static final String PREF_CHECK_REGIONS = PREFIX + ".checkPresetsRegions";
 
     /**
      * The preference key for source files
@@ -159,6 +163,7 @@ public class TagChecker extends TagTest implements TaggingPresetListener {
      * The preference key to search for presets - used before upload
      */
     public static final String PREF_CHECK_PRESETS_TYPES_BEFORE_UPLOAD = PREF_CHECK_PRESETS_TYPES + BEFORE_UPLOAD;
+    public static final String PREF_CHECK_REGIONS_BEFORE_UPLOAD = PREF_CHECK_REGIONS + BEFORE_UPLOAD;
 
     /**
      * The preference key for the list of tag keys that are allowed to be the same on a multipolygon and an outer way
@@ -175,18 +180,21 @@ public class TagChecker extends TagTest implements TaggingPresetListener {
     protected boolean checkComplex;
     protected boolean checkFixmes;
     protected boolean checkPresetsTypes;
+    protected boolean checkRegions;
 
     protected JCheckBox prefCheckKeys;
     protected JCheckBox prefCheckValues;
     protected JCheckBox prefCheckComplex;
     protected JCheckBox prefCheckFixmes;
     protected JCheckBox prefCheckPresetsTypes;
+    protected JCheckBox prefCheckRegions;
 
     protected JCheckBox prefCheckKeysBeforeUpload;
     protected JCheckBox prefCheckValuesBeforeUpload;
     protected JCheckBox prefCheckComplexBeforeUpload;
     protected JCheckBox prefCheckFixmesBeforeUpload;
     protected JCheckBox prefCheckPresetsTypesBeforeUpload;
+    protected JCheckBox prefCheckRegionsBeforeUpload;
 
     // CHECKSTYLE.OFF: SingleSpaceSeparator
     protected static final int EMPTY_VALUES                     = 1200;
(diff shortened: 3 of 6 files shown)


src/org/openstreetmap/josm/gui/tagging/presets/TaggingPreset.java
```diff
@@ -4,6 +4,7 @@
 import static org.openstreetmap.josm.tools.I18n.tr;
 import static org.openstreetmap.josm.tools.I18n.trc;
 import static org.openstreetmap.josm.tools.I18n.trn;
+import static org.openstreetmap.josm.tools.Territories.isIso3166Code;
```
tsmock, line 7: You aren't actually using this right now. But you probably will be.

src/org/openstreetmap/josm/gui/tagging/presets/TaggingPreset.java
```diff
@@ -39,16 +40,7 @@
...
-import org.openstreetmap.josm.data.osm.IPrimitive;
-import org.openstreetmap.josm.data.osm.OsmData;
-import org.openstreetmap.josm.data.osm.OsmDataManager;
-import org.openstreetmap.josm.data.osm.OsmPrimitive;
-import org.openstreetmap.josm.data.osm.Relation;
-import org.openstreetmap.josm.data.osm.RelationMember;
-import org.openstreetmap.josm.data.osm.Tag;
-import org.openstreetmap.josm.data.osm.Tagged;
-import org.openstreetmap.josm.data.osm.Way;
+import org.openstreetmap.josm.data.osm.*;
```
tsmock, line 43: Quick comment: Don't use `*` imports. In IntelliJ Settings, go to `Editor` -> `Code Style` -> `Java` -> `Imports` and change `Class count to use import with '*'` and `Names count to use static import with '*'` to `999`.

While you are there, if you end up optimizing imports sometime, you should also change the `Import Layout` to be 
```
import static all other imports
<blank line>
import java.*
<blank line>
import javax.*
<blank line>
import all other imports
```

That really just minimizes the diff. It isn't wholly necessary, but it does avoid unnecessary code churn.

Running `ant pmd checkstyle` will find most of the code style issues.

```suggestion
import org.openstreetmap.josm.data.osm.DataSet;
import org.openstreetmap.josm.data.osm.IPrimitive;
import org.openstreetmap.josm.data.osm.OsmData;
import org.openstreetmap.josm.data.osm.OsmDataManager;
import org.openstreetmap.josm.data.osm.OsmPrimitive;
import org.openstreetmap.josm.data.osm.Relation;
import org.openstreetmap.josm.data.osm.RelationMember;
import org.openstreetmap.josm.data.osm.Tag;
import org.openstreetmap.josm.data.osm.Tagged;
import org.openstreetmap.josm.data.osm.Way;
```

src/org/openstreetmap/josm/gui/tagging/presets/TaggingPreset.java
```diff
@@ -141,6 +133,10 @@ public class TaggingPreset extends AbstractAction implements ActiveLayerChangeLi
      * The types as preparsed collection.
      */
     public transient Set<TaggingPresetType> types;
+    /**
+     * list of regions the preset is applicable for
+     */
+    Set<String> regions;
```
tsmock, line 139:
```suggestion
    private Set<String> regions;
```

We want to use a `setRegions` here, since it will let us take a string and split it. If this is _not_ private, then the reflection code will try to use this field instead of the `setRegions` method.

src/org/openstreetmap/josm/gui/tagging/presets/TaggingPreset.java
```diff
@@ -605,6 +601,25 @@ public int showDialog(Collection<OsmPrimitive> sel, boolean showNewRelation) {
 
         int answer = 1;
         boolean canCreateRelation = types == null || types.contains(TaggingPresetType.RELATION);
+        if(regions!=null && !sel.isEmpty()){
```
tsmock, line 604:
```suggestion
        if (Utils.isEmpty(this.regions)) {
```
We have a specific utility method for checking if a collection is `null` or `isEmpty`.

src/org/openstreetmap/josm/gui/tagging/presets/TaggingPreset.java
```diff
@@ -605,6 +601,25 @@ public int showDialog(Collection<OsmPrimitive> sel, boolean showNewRelation) {
...
+            boolean flag=true;
+            for(OsmPrimitive osm : sel){
+                for(String region : regions){
+
+                    //TODO iterate through the nodes of way and relation
+                    /*if(!isIso3166Code(region, osm.getCoor())){
```
tsmock, line 610: For this, I would look at how `isIso3166Code` is used elsewhere. If you right-click on the method name, one option is `Find Usages`. Use that and see how `Territories.isIso3166Code` is used elsewhere.

src/org/openstreetmap/josm/gui/tagging/presets/TaggingPreset.java
```diff
@@ -605,6 +601,25 @@ public int showDialog(Collection<OsmPrimitive> sel, boolean showNewRelation) {
...
+                        flag=false;
+                    }*/
+                }
+            }
+            if(flag){
+                new Notification(
+                        tr("The preset <i>{0}</i> cannot be applied to this region!", getLocaleName()))
+                        .setIcon(JOptionPane.WARNING_MESSAGE)
+                        .show();
+                return DIALOG_ANSWER_CANCEL;
```
tsmock, line 620: As was discussed on the call, we _want_ to allow users to apply presets that may not be valid outside of a specific region, but we do want to notify them of that. Returning `DIALOG_ANSWER_CANCEL` probably isn't the best decision for this.

src/org/openstreetmap/josm/tools/XmlObjectParser.java
```diff
@@ -136,6 +136,13 @@ private void setValue(Entry entry, String fieldName, String value0) throws SAXEx
             try {
                 Object c = current.peek();
                 Field f = entry.getField(fieldName);
+                if ("regions".equals(fieldName)) {
+                    List<String> regions = null;
+                    if (value != null) {
+                        regions = Arrays.asList(value.split(","));
+                    }
+                    entry.getField(fieldName).set(current.peek(), regions);
+                }
```
tsmock, line 145:
```suggestion
```
This is unnecessary. This can be done _entirely_ in `TaggingPreset` by creating a new method `setRegions(String regions)`, e.g.

```java
    /**
     * Get the regions for the preset
     * @return The regions that the preset is valid for
     * @apiNote This is not {@code getRegions} just in case we decide to make {@link TaggingPreset} a record class.
     * @since xxx
     */
    public final Set<String> regions() {
        return this.regions;
    }

    /**
     * Set the regions for the preset
     * @param regions The region list (comma delimited)
     * @since xxx
     */
    public final void setRegions(String regions) {
        this.regions = Collections.unmodifiableSet(new HashSet<>(Arrays.asList(regions.split(","))));
    }
```

Tip: You might want to ensure that the region(s) are valid. Use `Territories.getKnownIso3166Codes()`. Also, you _should_ intern the country code strings (`Utils.intern`).

resources/data/defaultpresets.xml
```diff
@@ -2844,7 +2844,7 @@
             <reference ref="link_contact_address_payment" />
             <preset_link preset_name="Building" />
         </item> <!-- Fuel -->
-        <item name="Charging Station" icon="presets/vehicle/charging_station.svg" type="node" preset_name_label="true">
+        <item name="Charging Station" icon="presets/vehicle/charging_station.svg" regions="US,CA" type="node" preset_name_label="true">
```
tsmock, line 2847: This isn't actually limited to the United States and Canada. We'll have to remember to drop this prior to merge.
```suggestion
        <item name="Charging Station" icon="presets/vehicle/charging_station.svg" type="node" preset_name_label="true">
```

tsmock (changes requested):

Sarabjeet108:
@tsmock 

I initially was working on validating every node of a relation, but I found the test class _Address.java_ helpful, which contains a method that validates using the center of a _ways_ and _relations_, as you explained on the call. So I used the same approach for validation. 

Only an error message is displayed if the primitive is being mapped outside a given region.

src/org/openstreetmap/josm/gui/tagging/presets/TaggingPreset.java
Sarabjeet108: Thank you for the tip. 

The _checkstyle_ command is really helpful :)

resources/data/defaultpresets.xml
Woazboat: You could (/should) just change it locally without committing the change to git

Sarabjeet108: That seems like the better option.

Sorry, will keep that in mind.

src/org/openstreetmap/josm/gui/tagging/presets/TaggingPreset.java
tsmock: No problem. It is something that I frequently have to run (and sometimes I run, change something, and forget to run it again...)

Sarabjeet108:
@tsmock 

What more work needs to be done on this?

As per the changes, a warning message is displayed if a primitive is marked outside the given region.

resources/data/defaultpresets.xml
simonpoole: The easy solution for this is to use a custom preset for testing / dev purposes and simply disable the default preset.

simonpoole:
General comment: I don't know if JOSM has a policy on this, but I do have one of always rebasing PR commits on the HEAD of the main dev branch (in this case master). It really makes everything a lot simpler and avoids conflicts down the road.

simonpoole:
UI suggestion: given that search is returning all presets I would suggest that the region it is applicable to should be displayed in the preset modal as the behaviour is now, you don't know till -after- you've applied the preset that it is for a different regions.

Potentially the search behaviour might be controlled by a preference that governs if out of region presets are returned or not, or alternatively ask the user if "out of region" results should be returned if there are no "in region" results.

tsmock:
> What more work needs to be done on this?

I'll take a look at it again. I'm (technically) on vacation right now, so responses might be infrequent. But I'll pull down your current work once I finish going through the JOSM release process (up to pulling i18n updates -- I won't be doing a release until Friday, just in case there are some late-breaking bugs).

> I don't know if JOSM has a policy on this, but I do have one of always rebasing PR commits on the HEAD of the main dev branch (in this case master). It really makes everything a lot simpler and avoids conflicts down the road.

We don't have a policy on PRs, since I just take the PR in patch form and apply it (effectively squashing the commits). But this _is_ a good policy in general; for projects which are developed with git, I typically try to rebase my PRs weekly (and sometimes more frequently if it seems like the maintainers will be merging it soon).

tsmock:
> given that search is returning all presets I would suggest that the region it is applicable to should be displayed in the preset modal

This can be done by doing the following:
1. Extending `TaggingPresetValidation.java` to run the `TagChecker` class.
2. Modify `TagChecker.java` to check the regions

The end result will look something like this:
image

We should probably look into having a section in the dialog for warnings and/or show the warnings to the user after they apply the preset, _but that should be a different project_.

If you do 1 and 2, then you can (and should) get rid of the `if (!Utils.isEmpty(this.regions))` block.

Sarabjeet108:
@tsmock could you please review the commits for the Validation class? Specifically, handling the error for regions.

src/org/openstreetmap/josm/data/validation/tests/TagChecker.java
```diff
@@ -687,6 +696,33 @@ public void check(OsmPrimitive p) {
...
             }
         }
+
+        if(checkRegions) {
+            TaggingPreset preset = new TaggingPreset();
+            if(preset.regions()!= null) {
+                LatLon center = null;
+                if (p instanceof Node) {
+                    center = ((Node) p).getCoor();
+                } else if ((p instanceof Way) || (p instanceof Relation)) {
```
tsmock, line 706:
```suggestion
                } else {
```
All other possibilities have `getBBox`, which is why you didn't have to cast in the next statement.

tsmock (changes requested):

Sarabjeet108:
@tsmock Could you please review this? Is there anything else that needs to be done?

tsmock:
@Sarabjeet108 : I _believe_ you could have re-requested review under the `Reviewers` section in the right sidebar of the PR page.

src/org/openstreetmap/josm/data/validation/tests/TagChecker.java
```diff
@@ -654,15 +663,21 @@ public void check(OsmPrimitive p) {
         checkMultipolygonTags(p);
         }
 
+        final Collection<TaggingPreset> matchingPresets;
+        TagMap tags = p.getKeys();
+        if (checkPresetsTypes || checkRegions) {
+            matchingPresets = presetIndex.entrySet().stream()
+                    .filter(e -> TaggingPresetItem.matches(e.getValue(), tags))
+                    .map(Entry::getKey)
+                    .collect(Collectors.toCollection(LinkedHashSet::new));
```
tsmock, line 672: I know you just copied and pasted, but have you considered _why_ `LinkedHashSet` is used instead of the more standard `ArrayList`?

src/org/openstreetmap/josm/data/validation/tests/TagChecker.java
```diff
@@ -654,15 +663,21 @@ public void check(OsmPrimitive p) {
         checkMultipolygonTags(p);
         }
 
+        final Collection<TaggingPreset> matchingPresets;
+        TagMap tags = p.getKeys();
```
tsmock, line 667: I'd put the `tags` definition inside the `if` statement, much like `matchingPresets`.

tsmock (changes requested):
You still need to update the XSD. See https://github.com/JOSM/josm/blob/master/resources/data/tagging-preset.xsd . This is used to validate presets.

Other than that, I think this is "ok". I would have done some other things (for example, are some _values_ for a preset region specific?).

src/org/openstreetmap/josm/data/validation/tests/TagChecker.java
Sarabjeet108: One thing I know about `HashSet` is that it doesn't allow duplicate entries, which is not the case with `ArrayList`.

I had to search for `LinkedHashSet`. It maintains the order of the inserted elements. Is this the reason why `LinkedHashSet` has been used?

simonpoole: You are correct in that
- List maintains insert order, but allows duplicates
- Set doesn't allow duplicates, but doesn't guarantee that insertion order will be maintained

LinkedHashSet does maintain insertion oder and deduplication, but that comes at the cost of internally being quite complex, so this should only be used when both qualities are actually required.

tsmock: I suspect that `LinkedHashSet` was used mostly for deduplication. It was initially added to the code in [r15788](https://josm.openstreetmap.de/changeset/15788/josm/).

I don't believe any of the code written depends upon the _order_ of that list, so either a standard `List` class (with `.distinct()` in the stream) will work _or_ a standard `Set` class will work.

In this case, it isn't terribly important since we aren't passing the set around, but it is in code that can be run millions of times in a loop (if someone decides to validate a country, for example).

IIRC, you are using IntelliJ. If you ever encounter code that you do not understand the purpose of, you can right-click in the "gutter" and enable annotations with `Annotate`. This will then let you click on a line in the gutter to see the commit information. You can also right-click on the line in the gutter to annotate a _previous_ revision, which can be really handy when someone decides to bulk edit stuff.

resources/data/tagging-preset.xsd
```diff
@@ -121,7 +121,7 @@
     <complexType name="item">
         <annotation>
             <documentation>
-                Every item is one annotation set to select from. name is required, type and preset_name_label are recommended, icon and name_template are optional attributes.
+                Every item is one annotation set to select from. name is required, type and preset_name_label are recommended and icon, name_template, regions and exclude_regions are optional attributes.
```
tsmock, line 124: Can you reread this sentence. I don't think it flows well. Maybe replace the first `and` with `. `.

tsmock (changes requested):

tsmock:
In [18918/josm](https://josm.openstreetmap.de/changeset/18918/josm):

Fix [#23290](https://josm.openstreetmap.de/ticket/23290): Validate the regions a tag is expected to be in (patch by Sarabjeet108, modified)

Modifications are as follows:

    Allow the use of the new region attributes for keys inside a preset
    Basic tests 

regions comes from Vespucci's extensions: [​https://vespucci.io/tutorials/presets/#extensions](https://vespucci.io/tutorials/presets/#extensions)

PR #123, opened 2023-07-20
Update search dialog

Sarabjeet108:
@tsmock

I discussed this with Simon about having a filter for the added attributes. We can also have this for region attributes, although I think it is unnecessary in the case of _deprecated_.

I am having the same issue as with the regions attribute, the validate method in TaggingPresetValidation class is not adding the errors from the _check_ method in TagChecker class.

### DIFF
--- a/src/org/openstreetmap/josm/gui/dialogs/SearchDialog.java
+++ b/src/org/openstreetmap/josm/gui/dialogs/SearchDialog.java
@@ -231,7 +231,7 @@ public class SearchDialog extends ExtendedDialog {
          * selected preset by the user. Every query is of the form ' group/sub-group/.../presetName'
          * if the corresponding group of the preset exists, otherwise it is simply ' presetName'.
          */
-        selector = new TaggingPresetSelector(false, false);
+        selector = new TaggingPresetSelector(false, false, false);
         selector.setBorder(BorderFactory.createTitledBorder(tr("Search by preset")));
         selector.setDblClickListener(ev -> setPresetDblClickListener(selector, editorComponent));
 
--- a/src/org/openstreetmap/josm/gui/tagging/presets/TaggingPresetSearchPrimitiveDialog.java
+++ b/src/org/openstreetmap/josm/gui/tagging/presets/TaggingPresetSearchPrimitiveDialog.java
@@ -67,7 +67,7 @@ public final class TaggingPresetSearchPrimitiveDialog extends ExtendedDialog {
         super(MainApplication.getMainFrame(), tr("Search for objects by preset"), tr("Search"), tr("Cancel"));
         setButtonIcons("dialogs/search", "cancel");
         configureContextsensitiveHelp("/Action/TaggingPresetSearchPrimitive", true /* show help button */);
-        selector = new TaggingPresetSelector(false, false);
+        selector = new TaggingPresetSelector(false, false, false);
         setContent(selector, false);
         selector.setDblClickListener(e -> buttonAction(0, null));
     }
--- a/src/org/openstreetmap/josm/gui/tagging/presets/TaggingPresetSelector.java
+++ b/src/org/openstreetmap/josm/gui/tagging/presets/TaggingPresetSelector.java
@@ -51,7 +51,7 @@ import org.openstreetmap.josm.tools.Destroyable;
 import org.openstreetmap.josm.tools.Utils;
 
 /**
- * GUI component to select tagging preset: the list with filter and two checkboxes
+ * GUI component to select tagging preset: the list with filter and three checkboxes
  * @since 6068
  */
 public class TaggingPresetSelector extends SearchTextResultListPanel<TaggingPreset>
@@ -66,9 +66,11 @@ public class TaggingPresetSelector extends SearchTextResultListPanel<TaggingPres
 
     private static final BooleanProperty SEARCH_IN_TAGS = new BooleanProperty("taggingpreset.dialog.search-in-tags", true);
     private static final BooleanProperty ONLY_APPLICABLE = new BooleanProperty("taggingpreset.dialog.only-applicable-to-selection", true);
+    private static final BooleanProperty DISPLAY_DEPRECATED = new BooleanProperty("taggingpreset.dialog.display-deprecated", true);
 
     private final JCheckBox ckOnlyApplicable;
     private final JCheckBox ckSearchInTags;
+    private final JCheckBox ckDeprecated;
     private final Set<TaggingPresetType> typesInSelection = EnumSet.noneOf(TaggingPresetType.class);
     private boolean typesInSelectionDirty = true;
     private final transient PresetClassifications classifications = new PresetClassifications();
@@ -199,8 +201,9 @@ public class TaggingPresetSelector extends SearchTextResultListPanel<TaggingPres
      * Constructs a new {@code TaggingPresetSelector}.
      * @param displayOnlyApplicable if {@code true} display "Show only applicable to selection" checkbox
      * @param displaySearchInTags if {@code true} display "Search in tags" checkbox
+     * @param displayDeprecated if {@code true} display "Applicable in region" checkbox
      */
-    public TaggingPresetSelector(boolean displayOnlyApplicable, boolean displaySearchInTags) {
+    public TaggingPresetSelector(boolean displayOnlyApplicable, boolean displaySearchInTags, boolean displayDeprecated) {
         super();
         lsResult.setCellRenderer(new ResultListCellRenderer());
         classifications.loadPresets(TaggingPresets.getTaggingPresets());
@@ -226,6 +229,15 @@ public class TaggingPresetSelector extends SearchTextResultListPanel<TaggingPres
             pnChecks.add(ckSearchInTags);
         } else {
             ckSearchInTags = null;
+        }
+
+        if (displayDeprecated) {
+            ckDeprecated = new JCheckBox();
+            ckDeprecated.setText(tr("Show deprecated tags"));
+            pnChecks.add(ckDeprecated);
+            ckDeprecated.addItemListener(e -> filterItems());
+        } else {
+            ckDeprecated = null;
         }
 
         add(pnChecks, BorderLayout.SOUTH);
@@ -254,11 +266,12 @@ public class TaggingPresetSelector extends SearchTextResultListPanel<TaggingPres
         String text = edSearchText.getText().toLowerCase(Locale.ENGLISH);
         boolean onlyApplicable = ckOnlyApplicable != null && ckOnlyApplicable.isSelected();
         boolean inTags = ckSearchInTags != null && ckSearchInTags.isSelected();
+        boolean isDeprecated = ckDeprecated != null && ckDeprecated.isSelected();
 
         DataSet ds = OsmDataManager.getInstance().getEditDataSet();
         Collection<OsmPrimitive> selected = (ds == null) ? Collections.<OsmPrimitive>emptyList() : ds.getSelected();
         final List<PresetClassification> result = classifications.getMatchingPresets(
-                text, onlyApplicable, inTags, getTypesInSelection(), selected);
+                text, onlyApplicable, inTags, isDeprecated, getTypesInSelection(), selected);
 
         final TaggingPreset oldPreset = getSelectedPreset();
         lsResultModel.setItems(Utils.transform(result, x -> x.preset));
@@ -279,7 +292,7 @@ public class TaggingPresetSelector extends SearchTextResultListPanel<TaggingPres
 
         private final List<PresetClassification> classifications = new ArrayList<>();
 
-        public List<PresetClassification> getMatchingPresets(String searchText, boolean onlyApplicable, boolean inTags,
+        public List<PresetClassification> getMatchingPresets(String searchText, boolean onlyApplicable, boolean inTags, boolean isDeprecated,
                 Set<TaggingPresetType> presetTypes, final Collection<? extends OsmPrimitive> selectedPrimitives) {
             final String[] groupWords;
             final String[] nameWords;
@@ -292,11 +305,11 @@ public class TaggingPresetSelector extends SearchTextResultListPanel<TaggingPres
                 nameWords = searchText.split("\\s", -1);
             }
 
-            return getMatchingPresets(groupWords, nameWords, onlyApplicable, inTags, presetTypes, selectedPrimitives);
+            return getMatchingPresets(groupWords, nameWords, onlyApplicable, inTags, isDeprecated, presetTypes, selectedPrimitives);
         }
 
-        public List<PresetClassification> getMatchingPresets(String[] groupWords, String[] nameWords, boolean onlyApplicable,
-                boolean inTags, Set<TaggingPresetType> presetTypes, final Collection<? extends OsmPrimitive> selectedPrimitives) {
+        public List<PresetClassification> getMatchingPresets(String[] groupWords, String[] nameWords, boolean onlyApplicable, boolean inTags,
+                 boolean isDeprecated, Set<TaggingPresetType> presetTypes, final Collection<? extends OsmPrimitive> selectedPrimitives) {
 
             final List<PresetClassification> result = new ArrayList<>();
             for (PresetClassification presetClassification : classifications) {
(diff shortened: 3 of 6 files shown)


src/org/openstreetmap/josm/gui/tagging/presets/TaggingPresetSelector.java
```diff
@@ -279,7 +292,7 @@ public static class PresetClassifications implements Iterable<PresetClassificati
 
         private final List<PresetClassification> classifications = new ArrayList<>();
 
-        public List<PresetClassification> getMatchingPresets(String searchText, boolean onlyApplicable, boolean inTags,
+        public List<PresetClassification> getMatchingPresets(String searchText, boolean onlyApplicable, boolean inTags, boolean isDeprecated,
```
tsmock, line 295: We should probably use an enum here. We could probably reuse the one I mentioned earlier (so `getMatchingPresets(String searchText, DisplayCheckboxes...`, so `DisplayCheckboxes` is probably the wrong name; see if you can find one yourself).

src/org/openstreetmap/josm/gui/tagging/presets/TaggingPresetSelector.java
```diff
@@ -292,11 +305,11 @@ public List<PresetClassification> getMatchingPresets(String searchText, boolean
...
             }
 
-            return getMatchingPresets(groupWords, nameWords, onlyApplicable, inTags, presetTypes, selectedPrimitives);
+            return getMatchingPresets(groupWords, nameWords, onlyApplicable, inTags, isDeprecated, presetTypes, selectedPrimitives);
         }
 
-        public List<PresetClassification> getMatchingPresets(String[] groupWords, String[] nameWords, boolean onlyApplicable,
-                boolean inTags, Set<TaggingPresetType> presetTypes, final Collection<? extends OsmPrimitive> selectedPrimitives) {
+        public List<PresetClassification> getMatchingPresets(String[] groupWords, String[] nameWords, boolean onlyApplicable, boolean inTags,
+                 boolean isDeprecated, Set<TaggingPresetType> presetTypes, final Collection<? extends OsmPrimitive> selectedPrimitives) {
```
tsmock, line 312: I should probably mention (at this point) why I think we should start moving these booleans to an enum: easier readability. There are some other advantages, but `function(bool, bool, bool, bool, bool, bool)` start to get _really_ hard to wrap your mind around. Some IDEs will give hints as to what each bool is _supposed to be_, but not all IDEs will.

As a specific example, GitHub doesn't show what a boolean is supposed to be for, so the reviewer/editor has to know the function definition, or go find it.

tsmock (changes requested):

src/org/openstreetmap/josm/gui/tagging/presets/TaggingPresetSelector.java
```diff
@@ -199,8 +201,9 @@ public String toString() {
      * Constructs a new {@code TaggingPresetSelector}.
      * @param displayOnlyApplicable if {@code true} display "Show only applicable to selection" checkbox
      * @param displaySearchInTags if {@code true} display "Search in tags" checkbox
+     * @param displayDeprecated if {@code true} display "Applicable in region" checkbox
      */
-    public TaggingPresetSelector(boolean displayOnlyApplicable, boolean displaySearchInTags) {
+    public TaggingPresetSelector(boolean displayOnlyApplicable, boolean displaySearchInTags, boolean displayDeprecated) {
```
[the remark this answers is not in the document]
Sarabjeet108: Should we add a checkbox for regions as well?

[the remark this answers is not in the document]
tsmock: We could, yes. Realistically, the "primary" caller will probably just use `DisplayCheckboxes.values()` instead of hardcoding the current values.

src/org/openstreetmap/josm/gui/tagging/presets/TaggingPresetSelector.java
Sarabjeet108: how about `SearchStringFilters`?

tsmock: That would probably work. But we have other locations where we have search strings. It might be better to have some more identifying information. Although this really depends upon where we put the `enum` -- if it is in its own file, it will need a much better identifier, such as `PresetSearchFilter`. If it remains in the `TaggingPresetSelector` class, then that would be part of the identifier (`TaggingPresetSelector.SearchStringFilters.DEPRECATED`).

src/org/openstreetmap/josm/gui/tagging/presets/TaggingPresetSelector.java
```diff
@@ -254,11 +266,12 @@ protected synchronized void filterItems() {
         String text = edSearchText.getText().toLowerCase(Locale.ENGLISH);
         boolean onlyApplicable = ckOnlyApplicable != null && ckOnlyApplicable.isSelected();
         boolean inTags = ckSearchInTags != null && ckSearchInTags.isSelected();
+        boolean isDeprecated = ckDeprecated != null && ckDeprecated.isSelected();
```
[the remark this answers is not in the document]
Sarabjeet108: Should we create EnumMap inside the enum class or it would be better in the `TaggingPresetSelector` class?

src/org/openstreetmap/josm/gui/tagging/presets/TaggingPresetSelector.java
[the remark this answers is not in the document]
Sarabjeet108: > Realistically, the "primary" caller will probably just use `DisplayCheckboxes.values()` instead of hardcoding the current values.

Does that mean we would use the preference for the checkbox associated with the filter like `ONLY_APPLICABLE.getPref()` instead of using `bool` value? (getPref will return boolean value)

Sarabjeet108:
I am facing a few problems

1. How will we initialize the values for the filters in the enum map? As a temporary fix, I assigned `true` to all values using a static block.

2. You mentioned that the primary caller would call using the `.values()` method. How will that work if we need the boolean values associated with the filters since `.values()` will return an array of the enum constants?

src/org/openstreetmap/josm/gui/dialogs/SearchDialog.java
```diff
@@ -231,7 +232,7 @@ public boolean isValid() {
          * selected preset by the user. Every query is of the form ' group/sub-group/.../presetName'
          * if the corresponding group of the preset exists, otherwise it is simply ' presetName'.
          */
-        selector = new TaggingPresetSelector(false, false, false);
+        selector = new TaggingPresetSelector(PresetSearchFilter.values());
```
tsmock, line 235: This would be the equivalent of `new TaggingPresetSelector(true, true, true)`, which I don't _think_ you wanted. You probably wanted to just do `new TaggingPresetSelector()`.

src/org/openstreetmap/josm/gui/tagging/presets/PresetSearchFilter.java
```diff
@@ -0,0 +1,72 @@
...
+
+    /**
+     * Gets the preference for the filter
+     * @param filter the filter to get the preference for
+     * @return true if the filter is enabled, false otherwise
+     * @since xxx
+     */
+    public Boolean getPref(PresetSearchFilter filter) {
+        return filtersPreference.get(filter);
+    }
```
tsmock, line 47: When I said, "we should use an EnumMap", this is not what I was thinking. I'll go find one of the methods using this enum and see if I can hack together something there as an example.

Generally speaking, `enum` classes should not have a state that can change. There are exceptions, _but_ they are _exceptions_.

src/org/openstreetmap/josm/gui/tagging/presets/TaggingPresetSelector.java
```diff
@@ -64,13 +63,12 @@ public class TaggingPresetSelector extends SearchTextResultListPanel<TaggingPres
...
-    private static final BooleanProperty SEARCH_IN_TAGS = new BooleanProperty("taggingpreset.dialog.search-in-tags", true);
-    private static final BooleanProperty ONLY_APPLICABLE = new BooleanProperty("taggingpreset.dialog.only-applicable-to-selection", true);
-    private static final BooleanProperty DISPLAY_DEPRECATED = new BooleanProperty("taggingpreset.dialog.display-deprecated", true);
-
-    private final JCheckBox ckOnlyApplicable;
-    private final JCheckBox ckSearchInTags;
-    private final JCheckBox ckDeprecated;
+    PresetSearchFilter ONLY_APPLICABLE = PresetSearchFilter.ONLY_APPLICABLE;
+    PresetSearchFilter SEARCH_IN_TAGS = PresetSearchFilter.SEARCH_IN_TAGS;
+    PresetSearchFilter DEPRECATED_TAGS = PresetSearchFilter.DEPRECATED_TAGS;
```
tsmock, line 68: You shouldn't need/want to do this. Importing them statically would work, or you can just use the qualified name (`PresetSearchFilter.DEPREACATED_TAGS`). I only saw these used in a `switch` statement, where they aren't needed.

src/org/openstreetmap/josm/gui/tagging/presets/TaggingPresetSelector.java
```diff
@@ -212,32 +218,30 @@ public TaggingPresetSelector(boolean displayOnlyApplicable, boolean displaySearc
...
-            ckDeprecated.setText(tr("Show deprecated tags"));
-            pnChecks.add(ckDeprecated);
-            ckDeprecated.addItemListener(e -> filterItems());
-        } else {
-            ckDeprecated = null;
+        for (PresetSearchFilter option : options) {
+            switch(option) {
+                case ONLY_APPLICABLE: {
+                    ckOnlyApplicable = new JCheckBox();
+                    ckOnlyApplicable.setText(tr(PresetSearchFilter.ONLY_APPLICABLE.getText()));
```
tsmock, line 225: You could use `option.getText()` here instead of calling the qualified name.

We want to have a map of option -> checkboxes, so `EnumMap<PresetSearchFilter, JCheckBox>` would work. Example:

```java
final Map<PresetSearchFilter, JCheckBox> checkboxes = new EnumMap<PresetSearchFilter, JCheckBox>(PresetSearchFilter.class);
for (PresetSearchFilterOption option : options) {
    final JCheckBox box = new JCheckBox();
    box.setText(tr(option));
    box.addItemListener(e -> filterItems());
    pnChecks.add(box);
    switch (option) {
    case SEARCH_IN_TAGS:
        box.setSelected(PREF_SEARCH_IN_TAGS.get()); 
        break;
        // Do other enum values
    }
}
```

src/org/openstreetmap/josm/gui/tagging/presets/PresetSearchFilter.java
```diff
@@ -0,0 +1,72 @@
...
+    private final String translatedText;
+
+    /**
+     * Constructor for the PresetSearchFilter enum.
+     * Initializes an enum constant with its corresponding translated text.
+     *
+     * @param translatedText The translated text associated with the enum constant.
+     */
+    PresetSearchFilter(String translatedText) {
+        this.translatedText = translatedText;
```
tsmock, line 61: To avoid confusion, this should probably be `textToBeTranslated` or something; `translatedText` _implies_ it has already been translated.

src/org/openstreetmap/josm/gui/tagging/presets/TaggingPresetSearchPrimitiveDialog.java
```diff
@@ -67,7 +67,7 @@ public static synchronized TaggingPresetSearchPrimitiveDialog getInstance() {
         super(MainApplication.getMainFrame(), tr("Search for objects by preset"), tr("Search"), tr("Cancel"));
         setButtonIcons("dialogs/search", "cancel");
         configureContextsensitiveHelp("/Action/TaggingPresetSearchPrimitive", true /* show help button */);
-        selector = new TaggingPresetSelector(false, false, false);
+        selector = new TaggingPresetSelector(PresetSearchFilter.values());
```
tsmock, line 70: This would be the equivalent of new `TaggingPresetSelector(true, true, true)`, which I don't think you wanted. You probably wanted to just do new `TaggingPresetSelector()`.

src/org/openstreetmap/josm/gui/tagging/presets/PresetSearchFilter.java
```diff
@@ -0,0 +1,72 @@
...
+    }
+
+    /**
+     * Gets the preference for the filter
+     * @param filter the filter to get the preference for
+     * @return true if the filter is enabled, false otherwise
+     * @since xxx
+     */
+    public Boolean getPref(PresetSearchFilter filter) {
+        return filtersPreference.get(filter);
```
tsmock, line 46: This should be done via the standard `Config` interface _or_ one of the `Property` classes.

src/org/openstreetmap/josm/gui/tagging/presets/TaggingPresetSelector.java
```diff
@@ -264,14 +268,12 @@ public void actionPerformed(ActionEvent ae) {
...
-        boolean onlyApplicable = ckOnlyApplicable != null && ckOnlyApplicable.isSelected();
-        boolean inTags = ckSearchInTags != null && ckSearchInTags.isSelected();
-        boolean isDeprecated = ckDeprecated != null && ckDeprecated.isSelected();
 
         DataSet ds = OsmDataManager.getInstance().getEditDataSet();
         Collection<OsmPrimitive> selected = (ds == null) ? Collections.<OsmPrimitive>emptyList() : ds.getSelected();
         final List<PresetClassification> result = classifications.getMatchingPresets(
-                text, onlyApplicable, inTags, isDeprecated, getTypesInSelection(), selected);
+                text, ONLY_APPLICABLE.getPref(PresetSearchFilter.ONLY_APPLICABLE), SEARCH_IN_TAGS.getPref(PresetSearchFilter.SEARCH_IN_TAGS),
+                DEPRECATED_TAGS.getPref(PresetSearchFilter.DEPRECATED_TAGS), getTypesInSelection(), selected);
```
tsmock: Ideally, you would just use the same EnumMap from one of my other comments (`EnumMap<PresetSearchFilter, JCheckBox>`) instead.

Example: `map.entrySet().stream().filter(entry -> entry.getValue().isSelected().map(Map.Entry::getKey).toArray(PresetSearchFilter[]::new)`.

tsmock (changes requested):

src/org/openstreetmap/josm/gui/tagging/presets/TaggingPresetSelector.java
```diff
@@ -64,11 +65,15 @@ public class TaggingPresetSelector extends SearchTextResultListPanel<TaggingPres
...
 
-    private static final BooleanProperty SEARCH_IN_TAGS = new BooleanProperty("taggingpreset.dialog.search-in-tags", true);
-    private static final BooleanProperty ONLY_APPLICABLE = new BooleanProperty("taggingpreset.dialog.only-applicable-to-selection", true);
+    Map<PresetSearchFilter, Boolean> ckpreferences = new EnumMap<>(PresetSearchFilter.class);
+    {
+        //initial value of the checkboxes
+        ckpreferences.put(PresetSearchFilter.ONLY_APPLICABLE, true);
+        ckpreferences.put(PresetSearchFilter.SEARCH_IN_TAGS, true);
+        ckpreferences.put(PresetSearchFilter.DEPRECATED_TAGS, false);
+    }
```
tsmock, line 74: We should probably keep the properties.

Realistically, the map can be populated with `ckpreferences.computeIfAbsent(key, k -> PROPERTY.get())`.

Using (and saving) the preferences lets us keep the checkbox state between application restarts.

src/org/openstreetmap/josm/gui/tagging/presets/PresetSearchFilter.java
```diff
@@ -0,0 +1,37 @@
+// License: GPL. For details, see LICENSE file.
+package org.openstreetmap.josm.gui.tagging.presets;
+
+import static org.openstreetmap.josm.tools.I18n.marktr;
+
+/**
+ * This enum defines different filters for searching presets.
+ */
```
tsmock, line 8:
```suggestion
 * @since xxx
 */
```

tsmock (changes requested):

PR #125, opened 2023-08-07
JOSM #23105 - Create action to select shared child objects of selection

Woazboat:
https://josm.openstreetmap.de/ticket/23105

Create an action that selects the intersection of all child elements of the current selection

### DIFF
--- a/src/org/openstreetmap/josm/actions/SelectSharedChildObjectsAction.java
+++ b/src/org/openstreetmap/josm/actions/SelectSharedChildObjectsAction.java
@@ -1,0 +1,68 @@
+// License: GPL. For details, see LICENSE file.
+package org.openstreetmap.josm.actions;
+
+import static org.openstreetmap.josm.gui.help.HelpUtil.ht;
+import static org.openstreetmap.josm.tools.I18n.tr;
+
+import java.awt.event.ActionEvent;
+import java.awt.event.KeyEvent;
+import java.util.Collection;
+import java.util.Collections;
+import java.util.HashSet;
+import java.util.Set;
+
+import org.openstreetmap.josm.data.osm.IPrimitive;
+import org.openstreetmap.josm.data.osm.OsmPrimitive;
+import org.openstreetmap.josm.tools.Shortcut;
+import org.openstreetmap.josm.tools.Utils;
+
+/**
+ * Select child objects (way nodes and relation members) that are shared by all objects in the current selection.
+ */
+public class SelectSharedChildObjectsAction extends JosmAction {
+
+    /**
+     * Create a new SelectSharedChildObjectsAction
+     */
+    public SelectSharedChildObjectsAction() {
+        super(tr("Shared Child Objects"),
+                null,
+                tr("Select child objects (way nodes and relation members) that are shared by all objects in the current selection"),
+                Shortcut.registerShortcut("selection:sharedchildobjects",
+                    tr("Selection: {0}", tr("Shared Child Objects")),
+                    KeyEvent.CHAR_UNDEFINED, Shortcut.NONE),
+                true);
+        setHelpId(ht("/Action/SelectSharedChildObjectsAction"));
+    }
+
+    @Override
+    public void actionPerformed(ActionEvent e) {
+        Collection<OsmPrimitive> selection = getLayerManager().getActiveDataSet().getSelected();
+        Set<? extends IPrimitive> shared = getSharedChildren(selection);
+        getLayerManager().getActiveDataSet().setSelected(shared);
+    }
+
+    private Set<? extends IPrimitive> getSharedChildren(Collection<? extends IPrimitive> selection) {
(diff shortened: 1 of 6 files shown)


src/org/openstreetmap/josm/actions/SelectSharedChildObjectsAction.java
```diff
@@ -0,0 +1,68 @@
...
+import java.util.Set;
+
+import org.openstreetmap.josm.data.osm.IPrimitive;
+import org.openstreetmap.josm.data.osm.OsmPrimitive;
+import org.openstreetmap.josm.tools.Shortcut;
+import org.openstreetmap.josm.tools.Utils;
+
+/**
+ * Select child objects (way nodes and relation members) that are shared by all objects in the current selection.
+ */
```
tsmock, line 21:
```suggestion
 * @since xxx
 */
```

src/org/openstreetmap/josm/actions/SelectSharedChildObjectsAction.java
```diff
@@ -0,0 +1,68 @@
...
+
+    @Override
+    public void actionPerformed(ActionEvent e) {
+        Collection<OsmPrimitive> selection = getLayerManager().getActiveDataSet().getSelected();
+        Set<? extends IPrimitive> shared = getSharedChildren(selection);
+        getLayerManager().getActiveDataSet().setSelected(shared);
+    }
+
+    private Set<? extends IPrimitive> getSharedChildren(Collection<? extends IPrimitive> selection) {
+        Set<? extends IPrimitive> sharedChildObjects = new HashSet<>(selection.stream()
```
tsmock, line 46: Thank you for using the base primitive interface. :)

src/org/openstreetmap/josm/data/osm/IPrimitive.java
```diff
@@ -545,4 +546,16 @@ static void resetPrimitiveChildren(IPrimitive p) {
...
+     * Get child primitives that are referred by this primitive.
+     * {@link Relation}: Members of the relation
+     * {@link Way}: Nodes used by the way
+     * {@link Node}: None
+     * @return List of child primitives
+     * @since xxx
+     */
+    default List<? extends IPrimitive> getChildren() {
+        return Collections.emptyList();
+    }
```
tsmock, line 560: I'm on the fence about having a `default` method.

An alternative would be to have a "sane" default method, e.g.
```suggestion
    default List<? extends IPrimitive> getChildren() {
        if (this instanceof IRelation) {
            return this.getMemberPrimitivesList();
        } else if (this instanceof IWay) {
            return this.getNodes();
        }
        return Collections.emptyList();
    }
```

Alternatively, we could override this method in the respective `INode`, `IWay`, and `IRelation` classes.

I don't think this is wholly necessary, but I think there might be a bunch of duplicated code where that is what is happening.

src/org/openstreetmap/josm/data/osm/Relation.java
```diff
@@ -426,6 +426,11 @@ public List<OsmPrimitive> getMemberPrimitivesList() {
         return Utils.transform(getMembers(), RelationMember::getMember);
     }
 
+    @Override
+    public List<OsmPrimitive> getChildren() {
+        return getMemberPrimitivesList();
+    }
+
```
tsmock, line 433: See comment in `IPrimitive`. It should either be there or `IRelation`. If you put it in `IRelation`. In `IRelation`, it might be a bit tricky to have it return the appropriate primitive class without breaking backwards compatibility, so it _might_ be OK there.

src/org/openstreetmap/josm/data/osm/Way.java
```diff
@@ -44,6 +44,11 @@ public List<Node> getNodes() {
         return new CopyList<>(nodes);
     }
 
+    @Override
+    public List<Node> getChildren() {
+        return getNodes();
+    }
+
```
tsmock, line 51: See comment in IPrimitive. It should either be there or IWay. Putting it in IWay is easy; `public List<N> getChildren() { return getNodes(); }`

src/org/openstreetmap/josm/actions/SelectSharedChildObjectsAction.java
```diff
@@ -0,0 +1,68 @@
...
+    /**
+     * Create a new SelectSharedChildObjectsAction
+     */
+    public SelectSharedChildObjectsAction() {
+        super(tr("Shared Child Objects"),
+                null,
+                tr("Select child objects (way nodes and relation members) that are shared by all objects in the current selection"),
+                Shortcut.registerShortcut("selection:sharedchildobjects",
+                    tr("Selection: {0}", tr("Shared Child Objects")),
+                    KeyEvent.CHAR_UNDEFINED, Shortcut.NONE),
```
tsmock, line 33: No default shortcut works for now. Especially since there aren't a lot of valid "easy" remaining keys for it.

tsmock (changes requested):

PR #126, opened 2023-08-10
Add support for alternative tagging of the presets.

Sarabjeet108:
Would it be better if we create two separate `List` for storing `PresetLink` objects with alternative="true" and the ones that don't have an alternative tag like presets under the "Edit also" section? 

The current implementation depends on the order in which `preset_link` is defined in the XML file. It does not work if the preset_link with the alternative attribute is defined before the non-alternative ones or between them.

### DIFF
--- a/src/org/openstreetmap/josm/gui/tagging/presets/TaggingPreset.java
+++ b/src/org/openstreetmap/josm/gui/tagging/presets/TaggingPreset.java
@@ -420,6 +420,7 @@ public class TaggingPreset extends AbstractAction implements ActiveLayerChangeLi
         };
         JPanel linkPanel = new JPanel(new GridBagLayout());
         TaggingPresetItem previous = null;
+        boolean alternativeLabel = false; // true if an alternative label has been added
         for (TaggingPresetItem i : data) {
             if (i instanceof Link) {
                 i.addToPanel(linkPanel, itemGuiSupport);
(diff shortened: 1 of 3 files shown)


tsmock:
It would probably be preferable to have two separate lists. We could sort based off of the `alternative` tag, but then we start getting into changing the order of the alternatives. This is probably undesirable, if the preset author put the _preferred_ alternative first.

Sarabjeet108:
I added lists to store the PresetLink objects. Is there a need to sort the tags in the list?

simonpoole:
In a follow up to @tsmock comment: from a functional pov an "alternative preset" is different than a "linked preset". The later offers additional tags that might be added, the former replaces the current tagging with an alternative.

tl;dr the alternative preset links need a separate heading/label that makes the above clear (the "Edit also..." is very confusing here).

src/org/openstreetmap/josm/gui/tagging/presets/TaggingPreset.java
```diff
@@ -419,23 +418,43 @@ public void applyComponentOrientation(ComponentOrientation o) {
...
+            }
+        }
+
+        if (!editAlsoTags.isEmpty()) {
+            PresetLink link = new PresetLink();
+            itemPanel.add(link.createLabel(), GBC.eol().insets(0, 8, 0, 0));
+            for (PresetLink links : editAlsoTags) {
+                links.addToPanel(itemPanel, itemGuiSupport);
+                p.hasElements = true;
+            }
```
tsmock, line 457: If you consolidate the `createLabel`/`createAlternativeLabel`, this could be done like so:
```suggestion
        for (List<PresetLink> linkList : Arrays.asList(alternativeTags, editAlsoTags)) {
            if (!linkList.isEmpty()) {
                PresetLink link = new PresetLink();
                itemPanel.add(link.createLabel(), GBC.eol().insets(0, 8, 0, 0));
                for (PresetLink links : linkList) {
                    links.addToPanel(itemPanel, itemGuiSupport);
                    p.hasElements = true;
                }
            }
        }
```

tsmock (changes requested):

src/org/openstreetmap/josm/gui/tagging/presets/TaggingPreset.java
Sarabjeet108: `link.isAlternative()` is always `false`, therefore same label is being created for both preset alternative and editalso tags. We may want to create label based on the `links` of the `linkedList`

PR #127, opened 2023-09-01
Improve Preset for Charging Stations

dbrgn:
Two years ago, the Swiss OSM community organized the "[Project of the Month: Charging Stations](https://wiki.openstreetmap.org/wiki/DE:Project_of_the_month_Switzerland/Charging_Stations)". For this mapping project, I created a custom preset for mapping Swiss EV charging stations: https://github.com/dbrgn/josm-presets I'm an EV driver myself, and I mapped probably hundreds of charging stations in Switzerland and other European countries.

That preset is quite specific to the Swiss market, so it's not applicable to all countries. However, based on the things we learned from the preset being used by a few people over that month, we can make a lot of improvements to the current JOSM default preset for charging stations.

The commits of this PR are nicely separated and commented, so that if one of the changes is controversial, it could be removed. However, I can also squash the commits if that is preferred.

The biggest change is the commit that changes the approach for mapping charging sockets. Here's the commit description:

> First of all, I changed the input type from "combo" to "text". There are charging stations with a dozen or more charging spots, so having fixed numbers 1-6 doesn't make sense.
> 
> Next, I looked at how common these sockets are. Here is the list based on taginfo:
> 
> More than 1000: type2 (40204), type2_combo (13356), chademo (10754), schuko (6178), type2_cable (5397), type1 (3174), tesla_supercharger (1295)
> 
> Less than 1000: type3 (761), type1_combo (664), cee_blue (413), sev1011_t23 (202), cee_red_16a (180), cee_red_32a (119), nema_5_15 (98), nema_5_20 (46), sev1011_t13 (38)
> 
> Based on this research, I removed all entries that had less than 1000 occurrences. The affected sockets are either slow household-type sockets, or old standards that are being replaced with newer standards (type1_combo, type3).
> 
> Schuko is a bit special. It's slow and I assume most usages are for bicycle charging stations. However, an overpass query in europe for nodes tagged with Schuko in combination with both "amenity=charging_station" and "motorcar=yes" still turned up over 1700 entries, so I decided to leave it in the list for now.
> 
> One entry that was previously missing is "type2_cable". The distinction between just the socket and a cable is important for people that don't have their own charging cable, and for people that want to charge type1 cars using an adapter.

Reducing the number of sockets (by removing obscure socket types) gives us enough screen space to also map the output power per socket.

Screenshot before and after:

image

### DIFF
--- a/resources/data/defaultpresets.xml
+++ b/resources/data/defaultpresets.xml
@@ -2857,10 +2857,6 @@
                 <check key="fee" text="Charging fee" default="on" delete_if_empty="true" />
                 <check key="parking:fee" text="Parking fee" default="" delete_if_empty="true" />
             </checkgroup>
-            <space />
-            <combo key="voltage" text="Voltage" values="230,400"/>
-            <combo key="amperage" text="Amperage" values="16,32" />
-            <space />
             <label text="Types of vehicles which can be charged:" />
             <checkgroup columns="3">
                 <check key="motorcar" text="Cars" default="on" delete_if_empty="true" />
@@ -2877,6 +2873,15 @@
             <text key="socket:chademo" text="CHAdeMO" default="" delete_if_empty="true" />
             <text key="socket:tesla_supercharger" text="Tesla Supercharger" default="" delete_if_empty="true" />
             <text key="socket:schuko" text="Schuko" default="" delete_if_empty="true" />
+            <space />
+            <label text="Charging Power (e.g. '22 kW'):" />
+            <text key="socket:type1:output" text="IEC Type 1" default="" delete_if_empty="true" />
+            <text key="socket:type2:output" text="IEC Type 2 (Mennekes) socket only" default="" delete_if_empty="true" />
+            <text key="socket:type2_cable:output" text="IEC Type 2 (Mennekes) with cable" default="" delete_if_empty="true" />
+            <text key="socket:type2_combo:output" text="IEC Type 2 Combo (CCS)" default="" delete_if_empty="true" />
+            <text key="socket:chademo:output" text="CHAdeMO" default="" delete_if_empty="true" />
+            <text key="socket:tesla_supercharger:output" text="Tesla Supercharger" default="" delete_if_empty="true" />
+            <text key="socket:schuko:output" text="Schuko" default="" delete_if_empty="true" />
             <space />
             <label text="Authentication:" />
             <checkgroup columns="4">
--- a/resources/data/defaultpresets.xml
+++ b/resources/data/defaultpresets.xml
@@ -2862,13 +2862,12 @@
             <combo key="amperage" text="Amperage" values="16,32" />
             <space />
             <label text="Types of vehicles which can be charged:" />
-            <checkgroup columns="4">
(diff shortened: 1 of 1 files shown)


ernstnaef:
In my opinion a big improvement over the old version. Thanks a lot!

What I would suggest to add is the possibiliy to inform about the floor / storey (key: level) in a parking house. 
Default being empty.
In case the charging station is on another level than 0 it would be helpful to find it.

tsmock:
I'll have to try and remember to go through this. For future reference we usually like to see a ticket on JOSM Trac with a link to the GitHub PR.

Anyway, a few notes (since I think I was the last one to touch that preset):
* Removing charging socket types for bicycles/scooters is fine, but I would like to see them moved to a more specific preset. Only problem is the icon (maybe just recolorize it from green to blue or add a bike?)
* Updates to the authentication methods are fine, with a caveat: "Most charging stations
nowadays [...]" indicates to me that we should probably keep `Phone call` and `Short message` for now

dbrgn:
Oh, sorry, I almost missed your replies!

> What I would suggest to add is the possibiliy to inform about the floor / storey (key: level) in a parking house.

Good idea, I'll add that!

> I'll have to try and remember to go through this. For future reference we usually like to see a ticket on JOSM Trac with a link to the GitHub PR.

I wasn't aware of that workflow. Maybe adding a [CONTRIBUTING file](https://github.blog/2012-09-17-contributing-guidelines/) would be a good idea, to point this out to potential contributors! I'll create a Trac ticket.

> Removing charging socket types for bicycles/scooters is fine, but I would like to see them moved to a more specific preset. Only problem is the icon (maybe just recolorize it from green to blue or add a bike?)

I don't have any experience with bike/scooter chargers... I think a preset should be done by someone who who has practical experience with these vehicles (e.g. typical sockets, etc), a bad preset is worse than no preset at all (because people will rely on it) 🙂

> Updates to the authentication methods are fine, with a caveat: "Most charging stations
nowadays [...]" indicates to me that we should probably keep Phone call and Short message for now

Hmm, let me rephrase that... I've visited hundreds of charging stations by now, and I've never seen one that must be activated by phone call or short message. Are you aware of regions where that is a valid authentication method (meaning that you identify yourself through the phone number)?

The only place where I could see this being used is "ask at the reception" types of chargers at hotels. However, that would first require checking in, so just putting the phone number there isn't really of any benefit.

dbrgn:
A row for entering the level was added.

I also opened a Trac ticket: https://josm.openstreetmap.de/ticket/23339

tsmock:
In [18919/josm](https://josm.openstreetmap.de/changeset/18919/josm):

Fix [#23339](https://josm.openstreetmap.de/ticket/23339): Improve preset for charging stations (patch by dbrgn)

    Generic fee -> fee + parking:fee (the two may be separate)
    capacity is now free-form text (some charging stations can have 20 or more chargers)
    covered has been added
    Remove non-vehicle transport methods like bicycle and scooter (preset is in the vehicle group)
    Unused socket types have been removed
    Replace voltage and amperage with max socket output power
    Update authentication methods
    Add level for when the charging station is in a multi-level parking garage

PR #86, opened 2022-01-27
Update handrail.svg

AntoninCurtit:
Proposal to update the [handrail](https://wiki.openstreetmap.org/wiki/Tag:barrier%3Dhandrail) icon.

image
image

PR #130, opened 2023-10-11
Update handrail.svg

AntoninDelFabbro:
Linked to #86—finally know how to do it.


tsmock:
Can you add `width="16" height="16"` to the svg header?

Did you draw it yourself, or did it come from another service? If the latter, please let us know so we can make certain to properly attribute the icon.

Also, we generally like to see tickets on [JOSM Trac](https://josm.openstreetmap.de/newticket) explaining why you think that the patch is necessary.

tsmock:
> Drawed myself, but you don't have to credit me.

I strongly prefer to credit contributors. I assume crediting it to `AntoninDelFabbro` will work (since I have to commit it via `svn`, which will lose the `git` authorship information).

> Okay, that's confusing but if you like so

We use our own infrastructure for keeping track of tickets and patches; this git repo is here mostly for contributors that like to use git. I do look at the PRs, but it having the reasoning for various patches/issues in the same place makes it easier to figure out why something was changed.

tsmock:
JOSM ticket: https://josm.openstreetmap.de/ticket/23232

tsmock:
In [18862/josm](https://josm.openstreetmap.de/changeset/18862/josm):

Fix [#23232](https://josm.openstreetmap.de/ticket/23232): Improve handrail.svg icon (drawn by AntoninDelFabbro, modified)

It is easier to distinguish this new icon from the icon for benches.

Modifications are as follows:

    Widen strokes to pixel sizes
    Ensure that vertical/horizontal strokes are within pixel boundaries

PR #132, opened 2023-10-13
Switch to api.openstreetmap.org API host

danieldegroot2:
Use `api.openstreetmap.org/api/` -and HTTPS- instead of `www.openstreetmap.org/api/*`.

(Is: https://github.com/openstreetmap/operations/issues/951)

Except in validating a (manual) url, where the HTTP version has been left as valid for now.
Optionally, warn users if they are using the `www` and/or HTTP url.

Also changes some other URLs which redirect from HTTP to HTTPS into HTTPS urls.

This GitHub PR does not have a linked JOSM issue ticket at https://josm.openstreetmap.de/report .
Commit descriptions are left as default 'updated file ...'.

cc: @tsmock

### DIFF
--- a/test/unit/org/openstreetmap/josm/JOSMFixture.java
+++ b/test/unit/org/openstreetmap/josm/JOSMFixture.java
@@ -154,8 +154,8 @@ public class JOSMFixture {
     }
 
     private static boolean isProductionApiUrl(String url) {
-        return url.startsWith("http://www.openstreetmap.org") || url.startsWith("http://api.openstreetmap.org")
-            || url.startsWith("https://www.openstreetmap.org") || url.startsWith("https://api.openstreetmap.org");
+        return url.startsWith("http://api.openstreetmap.org")
+            || url.startsWith("https://api.openstreetmap.org");
     }
 
     private void setupGUI() {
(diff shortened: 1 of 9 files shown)


test/unit/org/openstreetmap/josm/JOSMFixture.java
tsmock, line 1: This is a test file -- it has no effect in production. Probably safe, but I do need to run the tests, even though `JOSMFixture` should be rarely used at this point.

tsmock (changes requested):

test/unit/org/openstreetmap/josm/JOSMFixture.java
danieldegroot2: I can replace this PR by one with just the changes to `JOSMFixture.java` so you can start testing right away. Is this okay?
(Changes made on desktop browser version.)

tsmock: I can make the changes locally. It does take ~10 minutes for the tests to run. I'll do that as soon as I finish what I'm currently working on, if I don't see something urgent in the ticket tracker.

tsmock: The tests passed, so it looks like nothing in JOSM core needs the `www.` checks.

danieldegroot2: Let me know what you think should be done with the PR. You can always close it. ;-)

danieldegroot2:
Closing as redundant.

PR #135, opened 2024-01-05
Added more tags for aerialway=station

mcliquid:
Added more tags from https://wiki.openstreetmap.org/wiki/Tag:aerialway%3Dstation


gabortim:
The `text` fields should be `combo`s (dropdown lists) with values and appropriate descriptions.

PR #137, opened 2024-02-21
Added turning circle shape

guylamar2006:


GerdP:
Small correction: The change was in [18995/josm](https://josm.openstreetmap.de/changeset/18995/josm):

PR #139, opened 2024-05-27
Add armrest tag to amenity=bench preset

bompstable:
See https://josm.openstreetmap.de/ticket/23694


bompstable:
I've previously used a combo for this but after experimentation, I've decided I prefer to use a simple checkbox for this, but have set the text to "Armrests (one or more)" to make the usage clearer and to avoid text such as "Armrest(s)" which can be difficult to translate.

Here's what the current version using checkboxes looks like:
image

And here's the previous version using a combo:
image

matkoniecz:

PR #142, opened 2024-08-06
update default style to show conditional turn restrictions

andrewharvey:
https://wiki.openstreetmap.org/wiki/Key:restriction defines the restriction:conditional tag. This change ensures that restrictions tagged this way show up with an icon.

I used a transparent icon to distinguish these conditional restrictions from standard unconditional restrictions.

I was tempted to set the text based on the conditional `condition` however it appears that [JOSM's MapCSS implementation](https://josm.openstreetmap.de/browser/josm/trunk/src/org/openstreetmap/josm/gui/mappaint/mapcss/Functions.java) does not offer a function to obtain the key directly based on a regexp.

### DIFF
--- a/resources/styles/standard/elemstyles.mapcss
+++ b/resources/styles/standard/elemstyles.mapcss
@@ -131,6 +131,76 @@ node[restriction] {
     text: auto;
 }
 
+/*********************************/
+/* conditional turn restrictions */
+/*********************************/
+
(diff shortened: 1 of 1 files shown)


stoecker:
Applied.

andrewharvey:
Thanks for merging!

PR #144, opened 2024-09-29
[Issue 23945] do not zoom into entry if the entry has not changed

ireun:
Related to https://josm.openstreetmap.de/ticket/23945


tsmock:
In [19266/josm](https://josm.openstreetmap.de/changeset/19266/josm):

Fix [#23945](https://josm.openstreetmap.de/ticket/23945): ImageViewerDialog.displayImages() - do not re-center if the image has not changed (patch by IreuN)

PR #146, opened 2024-11-05
Update defaultpresets.xml to have a better descriptions of trail_visibility

felagund:
Implementing wiki here:

https://wiki.openstreetmap.org/wiki/Key:trail_visibility


matkoniecz:

PR #147, opened 2024-12-07
Update golf bunker from natural=sand to surface=sand

guylamar2006:
Golf bunker sand is manmade by golf course designer.


matkoniecz:

PR #148, opened 2024-12-07
Added grass type to golf driving range.

guylamar2006:


stoecker:
That's a MIRROR! Report anything in JOSM bugtracker. Anyway I see no sense in this change.

PR #149, opened 2024-12-17
JOSM #24041 - Add preset amenity=research_institute

SteveLz:
ticket [#24041](https://josm.openstreetmap.de/ticket/24041)
Wiki: https://wiki.openstreetmap.org/wiki/Tag:amenity%3Dresearch_institute


matkoniecz:

PR #150, opened 2024-12-22
Update SearchDialog.java to show hint for key is null

andrewharvey:
The current example of "key=" is not working for me for me and don't make a lot of sense. can a key even exist without a value? A more common scenario is searching for objects missing a key, which I've changed this hint to.


stoecker:
It's not "without a value", but with empty value. Sure than can exist, but probably not created by the common editors.

You example is simply "-key:".

Change rejected. P.S. This is a mirror. Tickets must be reported in JOSM bug tracker.

andrewharvey:
Archaic cases like where a key exists but without a value or an empty value, practically never happen and we shouldn't make it more confusing for users but showing syntax for these over the more common searches of "find where key exists" and find features "without key present". 

"key:" key exists
"!key:" key doesn not exist.

PR #153, opened 2025-02-27
#24098 done. Auto zoom mode added in SelectionListDialog.

Antisida:
Auto zoom mode added. Mode switcher added


matkoniecz:

PR #154, opened 2025-04-07
Accept 'expressway' as valid maxspeed source

ireun:
In Poland it was agreed to use `PL:expressway` as `source:maxspeed` value for S-class roads.

https://taginfo.geofabrik.de/europe:poland/tags/source%3Amaxspeed=PL%3Aexpressway
https://www.openstreetmap.org/changeset/159759046
https://github.com/Project-OSRM/osrm-backend/pull/7079


stoecker:
Added.

PR #155, opened 2025-05-06
Use `VolatileImage` for back-buffers to allow hardware-accelerated rendering

hfmkwi:
## Summary

This partially fixes Ticket [#20366][josm-20366] wherein rendering is abysmally slow, especially on HiDPI systems. Current behavior uses a double-buffer strategy with `BufferedImage`s, forcing virtually all rendering into software loops.[^1] The bespoke HiDPI buffer scaling with `AffineTransforms` compounds this performance degradation.[^2] Performance seems to suffer exponentially with each additional layer regardless of its data; imagery tile layers seem to contribute the most.[^3]

I’ve switched the buffers to `VolatileImage`s, so the renderer can now defer to hardware routines where possible. This also defers HiDPI-scaling to the hardware surface, eliminating the need for the existing software one. The performance improvement is most obvious when running JOSM with a hardware-accelerated pipeline and

1. panning an non-trivial OSM data layer with labels disabled;
2. panning an arbitrary number of enabled imagery layers;
3. working with area fill enabled; or
4. working at high zoom levels.

This cannot convert some inherently software-based render operations like

- `Graphics2D` canvas/text anti-aliasing;
- `GlyphVector`-based text rendering; and
- alpha compositing (and whatever else I couldn’t find).

These ought to be refactored to take advantage of the equivalent hardware routines, if possible. If not, then the user should have finer control over these (e.g., rendering hints).

## Miscellanea

This also fixes a couple of performance regressions:

- `MapView` was marked as a transparent component; and
- hovering over an OSM primitive would immediately invalidate the layer, even if inactive

[josm-20366]: <https://josm.openstreetmap.de/ticket/20366> "#20366 (JOSM rendering is very slow especially on UHD displays)"
[oracle-primitive-tracing]: <https://docs.oracle.com/en/java/javase/24/troubleshoot/java-2d.html#GUID-35400B5A-A20D-4AFE-8A2C-1E1BD8AA5687> "Primitive Tracing to Detect and Avoid Non-Accelerated Rendering"

[^1]: See [Oracle docs][oracle-primitive-tracing].
[^2]: Incidentally, this is probably why the recommendation to run with `-Dsun.java2d.uiScale=1` was a suggested solution, since the resulting `AffineTransform` would’ve been the identity matrix.
[^3]: Stacking imagery layers very quickly destroys the frame rate.

### DIFF
--- a/src/org/openstreetmap/josm/gui/MapView.java
+++ b/src/org/openstreetmap/josm/gui/MapView.java
@@ -7,7 +7,6 @@ import java.awt.AWTException;
 import java.awt.AlphaComposite;
 import java.awt.BasicStroke;
 import java.awt.Color;
-import java.awt.Component;
 import java.awt.Dimension;
 import java.awt.Graphics;
 import java.awt.Graphics2D;
@@ -333,22 +332,24 @@ LayerManager.LayerChangeListener, MainLayerManager.ActiveLayerChangeListener {
         return Arrays.asList(zoomSlider, scaler);
     }
 
-    private static VolatileImage getAcceleratedImage(Component mv, int width, int height) {
+    private VolatileImage getAcceleratedBuffer() {
         VolatileImage volatileImage;
+        var width = getWidth();
+        var height = getHeight();
         // Creating a VolatileImage is impossible if 1) we’re headless or 2) our component is isolated (i.e., not in a
         // container). The former is inherent to VolatileImage creation; the latter is not, so we need to check that to
         // produce optimal VolatileImages.
-        if (null != mv.getGraphicsConfiguration()) {
+        if (null != getGraphicsConfiguration()) {
             // TODO: allow user to toggle between SW-backed and HW-backed?
             var volatileImageCapabilities = new ImageCapabilities(true);
             try {
-                volatileImage = mv.getGraphicsConfiguration().createCompatibleVolatileImage(width, height, volatileImageCapabilities, Transparency.OPAQUE);
+                volatileImage = getGraphicsConfiguration().createCompatibleVolatileImage(width, height, volatileImageCapabilities, Transparency.OPAQUE);
             } catch (AWTException e) {
                 //
-                volatileImage = mv.getGraphicsConfiguration().createCompatibleVolatileImage(width, height, Transparency.OPAQUE);
+                volatileImage = getGraphicsConfiguration().createCompatibleVolatileImage(width, height, Transparency.OPAQUE);
             }
         } else {
-            volatileImage = mv.createVolatileImage(width, height);
+            volatileImage = createVolatileImage(width, height);
         }
         if (null != volatileImage) {
             volatileImage.setAccelerationPriority(1);
@@ -538,7 +539,7 @@ LayerManager.LayerChangeListener, MainLayerManager.ActiveLayerChangeListener {
                     || VolatileImage.IMAGE_INCOMPATIBLE == offscreenBuffer.validate(getGraphicsConfiguration())
                     || offscreenBuffer.getWidth() != getWidth()
                     || offscreenBuffer.getHeight() != getHeight()) {
-                offscreenBuffer = getAcceleratedImage(this, getWidth(), getHeight());
+                offscreenBuffer = getAcceleratedBuffer();
             }
             var g2 = offscreenBuffer.createGraphics();
             g2.setClip(g.getClip());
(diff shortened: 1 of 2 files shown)


src/org/openstreetmap/josm/gui/MapView.java
```diff
@@ -101,6 +103,9 @@ public class MapView extends NavigatableComponent
 implements PropertyChangeListener, PreferenceChangedListener,
 LayerManager.LayerChangeListener, MainLayerManager.ActiveLayerChangeListener {
 
+    private static final boolean IS_HEADLESS = GraphicsEnvironment.isHeadless();
+    private static final Rectangle VIRTUAL_DESKTOP_BOUNDS = getVirtualDesktopBounds();
```
stoecker, line 107: The virtual desktop is not static. It can change (i.e. add another monitor).

hfmkwi: This is an issue I thought about for a while, since there's no non-trivial way of solving it.[^1] There’s no public interface for listening to complete `GraphicsEnvironment`/`GraphicsDevice` changes.[^2]

I could have the virtual desktop dimensions recalculated on 
every `GraphicsConfiguration` change. This *might* work if

- `GraphicsEnvironment.getScreenDevices()` always returns a live view of current screen devices; and
- stale screen devices are immediately evicted alongside their `GraphicsConfigurations`s.

These were my initial assumption—but I can’t find any supporting documentation. In testing, the component GC is invalidated when it’s moved between screen devices. I think this is reasonable enough evidence for my assumption. For now, I can apply that solution, since it seems the best compromise.

[^1]: At least, there’s no solution I’ve found yet myself.
[^2]: I could use reflection to get the internal one, but then I’d be dealing with implementation details that would unnecessarily complicate this. I also can’t reliably test for these, and I doubt these are guaranteed compatibility with an arbitrary Java version.

stoecker: Assumptions sound sane. Go for it and we'll hope it works fine. While I appreciate the improvement I fear the issues which changes to such a central component can bring. Especially the multiscreen stuff was very fragile a long time.

hfmkwi: I get the concerns: I’ve approaching this with multi-head setups in mind since I’ve been testing on one. There are some hardware pipeline idiosyncracies that *might* affect render speeds depending on which screen device the application resides,[^1] but those seem platform- and pipeline-specific.[^2] Generally, this setup should be correct for both single- and multi-head setups. Internally, `SunVolatileImage` appears to use a GC-compatinke `BufferedImage` if something goes wrong with the hardware pipeline.

[^1]: On my machine, at least, the D3D pipeline runs on all screens and `VolatileImage`s are compatible only with their progenitor screen device, whereas the OGL one runs on display 0 only and the VIs appear to be resistant to that.
[^2]: Determining the actual renderer requires either a string comparison with screen device IDs or reflection to expose package-privage implementation details of the JDK.

stoecker: VIRTUAL_DESKTOP_BOUNDS indicates it is static and unchangeable (constant). I think virtualDesktopBounds is better.

stoecker: One checkstyle:
src/org/openstreetmap/josm/gui/MapView.java:369:43: Missing space after ','

stoecker: Ok. I did a first test on my non-HIDPI screen using a larger data set and nothing else and simply move it around. With your patch it feels significantly slower.

hfmkwi: > Ok. I did a first test on my non-HIDPI screen using a larger data set and nothing else and simply move it around. With your patch it feels significantly slower.

What render settings were your running with? Pipeline, anti-aliasing? Any sort of AA, text or canvas, is done entirely in software, and JFK docs caution against mixing hardware- and software-based rendering. Maybe I should have the AA preferences in the GUI warn about this?

hfmkwi: If these changes are sufficiently intrusive, I could also move this code to an entirely separate renderer that overrides paint methods as applicable.

stoecker: I used default settings (Linux openSUSE), no special setup.

I think the first thing we need is proper profiling code which also outputs the relevant parameters (pre-patch and post-patch), so that we can compare number instead of feelings. We can then use a screen recorder and playback to compare identical situations.

Than we can check where the code improves situation and where it makes it worse. Based on that we can either fix introduced bugs or decide to make different code for different setups.

stoecker:
Yes, Exactly that's what I mean. Add some "this took ..." statements in old and new code (in ms level). Then we can start to see what improved and what got worse. Don't do it too fine-grained (lots of work, little gain), but also not to simplified (doesn't help).

Also notice that profiling may influence the timing, so sometimes it's a good idea to first record the data and only print it later, as recording usually is fast, printing slow. Code can be ugly, as it gets removed afterwards, so static variables and the like are ok.

PR #163, opened 2026-05-06
Add Maven publish workflow for GitHub Actions

randybloxham5-arch:
This workflow builds a Maven package and publishes it to GitHub Packages upon release creation.


randybloxham5-arch:
Updated

randybloxham5-arch:
Updated

randybloxham5-arch:
Updated the maven files

stoecker:
First: @v4 for both actions is deprecated. Take v6 instead.

And second, well what was this about? Wrong project?

PR #164, opened 2026-05-22
Add building=semidetached_house to defaultpresets.xml

Rainer-o:
semidetached_house is used nearly 2 million times, so JOSM should have it in the preset's selection list: https://taginfo.openstreetmap.org/tags/building=semidetached_house


matkoniecz:

PR #167, opened 2026-08-06
Add fire_hydrant:diameter=500 to defaultpresets.xml

baltpeter:
I noticed that the validator was complaining about `500` being an unknown property value for `fire_hydrant:diameter`.

> Unknown property value - Value '500' for key 'fire_hydrant:diameter' is unknown, maybe one of [100, 200, 300, 400, 50] is meant? (2)

image

From looking at https://josm.openstreetmap.de/changeset/18352/josm and https://josm.openstreetmap.de/changeset/15465/josm, it seems like the way to fix that is to add the value to the default presets.


matkoniecz:

